PR #1, opened 2014-05-28
ltp-ddt sometimes cannot find scripts/basic/fixdep

ashleysommer:
Added the make_scripts task to the ltp-ddt bitbake recipe.

If no task has created the kernel scripts before ltp-ddt is run, there is a compile error complaining that it cannot find scripts/basic/fixdep. This patch ensures that these scripts are built before compiling starts.


shr-project:

PR #12, opened 2014-12-04
Add a new version of python psutil

gns-ank:
This is a simple addition of a new version of psutil. Should I create a pull request for other branches as well? I am not too familiar with the branch naming structure, but I would like to push it to dizzy as well as branches for releases scheduled after dizzy.

Note, that the python API between the old and this version are different.

Signed-off-by: Ignas Anikevicius <ignas.anikevicius@realvnc.com>


shr-project:

PR #16, opened 2015-09-16
add libgudev which is not provided by systemd anymore

afreof:
Systemd 225 does not provide gudev anymore. libgudev has been moved to a separate project. This patchset adds a bb for this new library. Further on networkmanager, modemmanager and libmbin depend now on libgudev instead of udev.


shr-project:

PR #19, opened 2015-12-04
switch from Google Code to Github for vim 7.4.481 in fido

andrewgoktepe:
Google Code is being decommissioned.  vim has already moved to Github and they have also changed their tag names to use dots instead of dashes. 

Intermittently for some reason, we are seeing yocto build failures with an inability to fetch the vim source from Google Code.  

I see that the openembedded master branch has already switched to github for the vim recipe.  This change is implementing that same migration, but for the fido branch retaining the same version of vim that was already in use in fido.


balister:
You should send this to the meta-oe least with the info requested by the
fido maintainer (Armin Kuster, I think)

We do not use pull requests on github.

Philip

andrewgoktepe:
Thanks, I just found this.  Will submit via official channels
http://www.openembedded.org/wiki/How_to_submit_a_patch_to_OpenEmbedded

balister:
Thanks.

PR #23, opened 2016-03-09
crash: add native package

parr0tr1ver:
Add crash-native package.

* append EXTRA_OEMAKE option to find ncurses library
* pass right install DESTDIR

Signed-off-by: Kai Kang <kai.kang@windriver.com>


shr-project:
We're not using github's pull requests, follow README and send this to ML

PR #27, opened 2016-08-24
meta-oe: Add iksemel 1.5 recipe

scthunderbolt:
iksemel 1.4 is blacklisted, iksemel 1.5 is needed to build


shr-project:
We're not using github's pull-requests, please follow the README and send it to openembedded-devel ML.

PR #29, opened 2016-10-06
jansson: update to 2.9

socec:
Signed-off-by: Igor Socec <igor.socec@pelagicore.com>

### DIFF
--- a/meta-oe/recipes-extended/jansson/jansson_2.9.bb
+++ b/meta-oe/recipes-extended/jansson/jansson_2.9.bb
@@ -1,12 +1,12 @@
 SUMMARY = "Jansson is a C library for encoding, decoding and manipulating JSON data"
 HOMEPAGE = "http://www.digip.org/jansson/"
 LICENSE = "MIT"
-LIC_FILES_CHKSUM = "file://LICENSE;md5=6946b728e700de875e60ebb453cc3a20"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=8b70213ec164c7bd876ec2120ba52f61"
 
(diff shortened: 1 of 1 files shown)


shr-project:
Read the Readme file and submit to openembedded-devel ML.

socec:
I have read the README file and it says the following:
You are encouraged to fork the mirror on github https://github.com/openembedded/meta-oe/ to share your patches, this is preferred for patch sets consisting of more than one patch. Other services like gitorious, repo.or.cz or self hosted setups are of course accepted as well, 'git fetch <remote>' works the same on all of them. We recommend github because it is free, easy to use, has been proven to be reliable and has a really good web GUI.

What is wrong with using Github then?

shr-project:
The 4 lines above it?

Send pull requests to openembedded-devel@lists.openembedded.org with '[meta-oe]' in the subject'

When sending single patches, please use something like:
'git send-email -M -1 --to openembedded-devel@lists.openembedded.org --subject-prefix=meta-oe][PATCH'

socec:
Then please remove the other lines that encourage me to fork the mirror on github to share patches.

(IMHO it is much nicer to handle pull requests here.)

graugans:
@socec I agree, it seems easier from the perspective of an typical github user to use Pull Requests instead of sending patches. The benefit of sending patches is, there are more eyes checking the patches and those projects are used to this work flow. 

@shr-project I wonder why you do not disable the possibility to create issues? Pull-Requests can not be disabled....

socec:
@graugans I agree with this workflow and I have sent patches to mailing lists before. My confusion comes from the fact that I initially opened the README to find the mailing list and then bumped into the encouragement lines about GitHub so decided to give it a go instead of sending the patch :)

PR #42, opened 2017-05-16
update tslib and xf86-input-tslib

merge:
Please note that these changes are __untested__ and done in a hurry. But the versions of these components you are using are very old. So either you take the change and test, or see this as a notification only :)

thanks


shr-project:

PR #43, opened 2017-05-22
audiofile: Inherit pkgconfig

tworaz:
Without this the build fails with the following error:

configure: line 16964: PKG_PROG_PKG_CONFIG: command not found
configure: line 16971: syntax error near unexpected token `FLAC,'
configure: line 16971: PKG_CHECK_MODULES(FLAC, flac >= 1.2.1, ac_cv_flac=yes, ac_cv_flac=no)'

Signed-off-by: Piotr Tworek <tworaz@tworaz.net>


shr-project:

PR #44, opened 2017-05-22
htop: Upgrade to version 2.0.2.

tworaz:
New version also needed small workaround for OEs non functional
ncurses-config script. Fortunately  pkg-config can be used instead.

Signed-off-by: Piotr Tworek <tworaz@tworaz.net>


shr-project:

PR #46, opened 2017-06-01
Backport from master

mnsgs:
Fix wrong checksum of devmem2


shr-project:

gizero:
So, shouldn't we drop (or at least reword) the README a couple of line below the one you highlight? To me it sounds like we are encouraging to use GitHub, at least for sets of more than one patch...

mnsgs:
Agree, this was exactly what mislead me.

shr-project:
Agreed, it's misleading. Patches to improve it are also welcome on openembedded-devel ML :).

PR #51, opened 2017-08-06
Fix uim problems

mickey-happygolucky:
Uim recipe is not maintained.


mickey-happygolucky:
I sent these patches to the ML.
close this.

Issue #50, opened 2017-07-28
Netperf SRC_URI link seems to be broken


helgeanl:
From what I can see in this [issue](https://github.com/HewlettPackard/netperf/issues/2) the ftp server for Netperf is closed down and all is moved to github. This causes the [netperf recipe](https://github.com/openembedded/meta-openembedded/blob/master/meta-networking/recipes-support/netperf/netperf_2.6.0.bb) in meta-networking to fail fetching.

shr-project:
we use github only as read-only mirror, very few people ever read issues here.

You should send patch to openembedded-devel ML as README suggests.

PR #52, opened 2017-08-27
fix broken SRC_URI

zeronote:
fix https://github.com/openembedded/meta-openembedded/issues/50


zeronote:
Just saw http://lists.openembedded.org/pipermail/openembedded-devel/2017-July/113762.html, closing

PR #53, opened 2017-08-29
plymouth: add conditional runtime dependency in libdrm

leiflm:
The DRM backend tries to `dlopen` `libdrm2.so` during runtime.

This patch also increases the package release to reflect that only
packaging changed.


shr-project:
Read:
https://www.openembedded.org/wiki/How_to_submit_a_patch_to_OpenEmbedded

and send the change to ML.

Also
1) you should use 4th PACKAGECONFIG parameter for runtime dependency
2) drop PR bump
3) add SOB line

leiflm:
Thanks for the feedback. It'll have to wait a bit then, since I'm behind a corporate firewall atm.
However, changes as per to your suggestion are already done.

PR #56, opened 2017-11-28
madplay support for meta-multimedia

vsivanagulu:
Working madplay recipe for yocto.


shr-project:
Patches should be sent to openembedded-devel as described in README.

PR #62, opened 2018-02-02
Patch 1

vsivanagulu:
added madplay patch for fido


shr-project:
As in your previous PR:
https://github.com/openembedded/meta-openembedded/pull/56

Patches should be sent to openembedded-devel as described in README.

PR #63, opened 2018-02-04
musicpd: update musicpd 0.20.16

MaxKellermann:
Also make lots of plugins optional.


kraj:
cdf33509f8c348a61cb65ab7b1336076991b319c upgraded to 0.20.18

PR #74, opened 2018-03-16
modemmanager: Update flag for disabling mbim

RPGillespie6:
Updated configure flag used for disabling mbim in modemmanager. The existing one does not work anymore.


kraj:
merged

PR #77, opened 2018-05-09
stm32flash: add recipe

g0hl1n:
Add recipe for version 0.5 of stm32flash, an open source flash program for
STM32 using the ST serial bootloader.

Signed-off-by: Richard Leitner <richard.leitner@skidata.com>


g0hl1n:
sent patch to openembedded-devel ML as stated in the README.

PR #81, opened 2018-05-29
Added the packages python-periphery, python-sijax and python-flask-sijax

FranzForstmayr:


kraj:
already applied. Please close it

PR #89, opened 2018-08-25
Copy breakpad .sym file in appropriate folder.

tramseyer:
Google breakpad requires a very specific folder structure which is only known after dump_syms has been executed.
Therefore the .sym file's first line has to be parsed in order to move the file where breakpad wants it.


kraj:
merged

PR #106, opened 2018-11-28
recipes-extended: add libzip 0.9

nsjodk:
libzip is a C library for reading, creating, and modifying zip
archives. Files can be added from data buffers, files, or
compressed data copied directly from other zip archives.
Changes made without closing the archive can be reverted.

Version 0.9 is a little dated and the newest version is 1.5.1.
However, there are some breaking API changes between versions,
so this older version is needed.

Signed-off-by: Niklas Johansson <niklas.johansson@prevas.dk>


kraj:
Can you post the patch to openembedded-devel mailing list please ?

kraj:
https://github.com/openembedded/meta-openembedded/commit/26ec99bfe70742bd8ab3c6642ce2fcf2d690de4e has already added version 1.5.1, what difference ?

nsjodk:
There are some breaking API changes between the newer version an this older version.

For some programs, it is non trivial to port to the newer version.

As a note. I've been in contact with the author and asked about the license. (Which is missing in the 0.9 archive). I therefore added the BSD license file he suggested, manually to the files folder.

Från: Khem Raj
Skickat: måndag 3 december 05:14
Ämne: Re: [openembedded/meta-openembedded] recipes-extended: add libzip 0.9 (#106)
Till: openembedded/meta-openembedded
Kopia: Niklas Johansson, Author


26ec99b<https://github.com/openembedded/meta-openembedded/commit/26ec99bfe70742bd8ab3c6642ce2fcf2d690de4e> has already added version 1.5.1, what difference ?

kraj:
In that case it should be done such that it can co-exist with 1.5.1, usually I would prefer to keep the version we already have and port the apps as needed. Maintaining old version is only done for already existing recipes.

nsjodk:
I thought the could co-exist.
However, I can see you point about maintaining old versions.
I will close this PR and move this recipe to another meta-layer where it is needed.
Thank you for you comments and feedback.

PR #108, opened 2018-12-07
Fix checksum for python-pyrex_0.9.9.bb

devcharted:


kraj:
this seems to be failing as well can you try to do -ccleanall and then build again ?

```
ERROR: python-pyrex-0.9.9-r4 do_fetch: Checksum failure fetching http://www.cosc.canterbury.ac.nz/greg.ewing/python/Pyrex/Pyrex-0.9.9.tar.gz
ERROR: python-pyrex-0.9.9-r4 do_fetch: Fetcher failure for URL: 'http://www.cosc.canterbury.ac.nz/greg.ewing/python/Pyrex/Pyrex-0.9.9.tar.gz'. Checksum mismatch!
File: '/mnt/a/yoe/downloads/Pyrex-0.9.9.tar.gz' has md5 checksum 515dee67d15d4393841e2d60e8341947 when 70ff83ad4207f07deb531745a71df813 was expected
File: '/mnt/a/yoe/downloads/Pyrex-0.9.9.tar.gz' has sha256 checksum 5f87df06831d0b3412eb4bc9d3fc2ee7bfae1b913d7da8c23ab2bf5699fb6b50 when 6a596d594b4b4b1ebfc09ccaa36732523eda12ca2acba05e1cbf3b12ad172107 was expected
If this change is expected (e.g. you have upgraded to a new version without updating the checksums) then you can use these lines within the recipe:
SRC_URI[md5sum] = "515dee67d15d4393841e2d60e8341947"
SRC_URI[sha256sum] = "5f87df06831d0b3412eb4bc9d3fc2ee7bfae1b913d7da8c23ab2bf5699fb6b50"
Otherwise you should retry the download and/or check with upstream to determine if the file has become corrupted or otherwise unexpectedly modified.
```

PR #111, opened 2019-01-07
[PATCH] libgpiod_1.2: add dependency on python when building python bindings

htot:
Commit ab54dd75 "libgpiod: Rrecommend python3 only for PN-python package" also removes the build dependency on python3 however this results in 
fatal error: Python.h: No such file or directory #include <Python.h>

Signed-off-by: Ferry Toth <ftoth@exalondelft.nl>


kraj:
applied c44679b4942a30ac89013343f987403ec63f0ece

htot:
Thanks

PR #113, opened 2019-01-16
Update the regex for PGMAJORVERSION to match v10+ clusters

adhameed:
Older versions of postgresql would generate MAJVER.MINVER in new clusters. 10+ do not. Simplify the regex to match PGMAJORVERSION from a single int.


kraj:
Please add your sign-off on patch

kraj:
fixed with a368821c96bc041a8d7f7dab7d4a653d59729d0e

PR #130, opened 2019-03-26
Add recipe for triggerhappy daemon

markubiak:
First time upstreaming a recipe, attempts were made to clean it up but please let me know if any style conventions are being violated.

Simple recipe for the 'triggerhappy' daemon, which runs shell commands in response to inputs from /dev/input. The debian release from the official repository was used, as it contains an initscript not contained in the master.

Points potentially requiring review:

- All daemons are, by default, ran as "nobody".
- My environment uses systemd, sysvinit script should probably be tested for anything debian-specific but on first glance it looks fine.
- When built with systemd (identified by pkg-config in the Makefile), the service is of type 'notify', requiring linkage to libsystemd. This is the intent of the PACKAGECONFIG section, works fine on my system.


markubiak:
Oops, based this off ``sumo``. I'll rebase to master and reopen.

PR #134, opened 2019-05-28
dhcpcd: Install to /sbin instead of /usr/sbin

ostaebler:
ifupdown expects dhcpcd to be in /sbin, so we install it in /sbin
instead of /usr/sbin.

Signed-off-by: Oliver Stäbler <oliver.staebler@bytesatwork.ch>


kraj:
Can we change ifupdown to use it from /usr/sbin instead ? since there might be more apps in OE expecting it in its original location, its better to change the consumer instead of source.

ostaebler:
Sorry for my (very) late answer.
Yes, this makes sense. I'll have another look at it.

PR #135, opened 2019-06-06
blueman_%.bbappend: Avoid PAK archive (application/x-pak)

persianpros:
Just check it in ubuntu you'll see.


kraj:
applied as 10c39b96a2

PR #136, opened 2019-06-10
gitver: Pass git directory argument to gitrev_run

hummerbliss:
${GITSHA} gives the following error otherwise
gitrev_run() missing 1 required positional argument: 'path'

Signed-off-by: Saikiran Madugula <hummerbliss@gmail.com>


kraj:
applied with 0e342ecaa4

PR #141, opened 2019-08-26
Don't hardcode path

persianpros:


meta-networking/recipes-connectivity/samba/samba_4.10.6.bb
```diff
@@ -146,7 +146,7 @@ do_install_append() {
     install -d ${D}${systemd_system_unitdir}
     install -m 0644 ${S}/bin/default/packaging/systemd/*.service ${D}${systemd_system_unitdir}/
     sed -e 's,\(ExecReload=\).*\(/kill\),\1${base_bindir}\2,' \
-        -e 's,/etc/sysconfig/samba,${sysconfdir}/default/samba,' \
+        -e 's,${sysconfdir}/sysconfig/samba,${sysconfdir}/default/samba,' \
```
schnitzeltony: This is wrong. In case ${sysconfdir} changes (that's what this patch is about) sed will stop working.

There are other cases like this in this patch

PR #146, opened 2019-10-08
Add zeus compatibility to all layer.conf files

palmer-dabbelt:
I have no idea if this is the right thing to do, but without the patch I
can't actually buil OE because none of these layers are compatible
with the change in openembedded-core to move to zeus.

Fixes: a5c9709b8d ("layer.conf: Update for zeus series") # openembedded-core

Signed-off-by: Palmer Dabbelt <palmer@dabbelt.com>


kraj:
merged with https://git.openembedded.org/meta-openembedded/commit/?id=fd1a0c9210b162ccb147e933984c755d32899efc

palmer-dabbelt:
Thanks.  Do you want these to go to some mailing list instead of here?

PR #161, opened 2019-12-11
libnftnl: upgrade 1.1.4 -> 1.1.5

tcunningham07:
Remove obselete patch:
    0001-Move-exports-before-symbol-definition.patch
Refresh patch:
    0002-avoid-naming-local-function-as-one-of-printf-family.patch
Add one new file in this patch:
    src/obj/synproxy.c

Signed-off-by: Todd Cunningham <tcunningham07@gmail.com>


tcunningham07:
Looks like I was overzealous with `snprintf` -> `snprintf_` in the patch refresh. I believe the latest change fixes the issue.

kraj:
this is merged now 9625a375eea655d02cb3d45ff98d56c821dda297

PR #171, opened 2020-01-17
Update cups-filters packaging according to oe-core

diegorondini:
As the crossscript is now fixed in oe-core:
http://git.openembedded.org/openembedded-core/commit/?id=2ce6ef29b9bb4f16ed9d78e166d455b7a6d968bf
adjust packaging accordingly by using libexecdir instead of libdir.
Additionally update cups-filters to latest version.


kraj:
merged with 413cb43ea9df8475920f464ab6db16fe2e68dbeb and fc0d5ab6e3d7c409be735fef4e58090ad79e177f

PR #190, opened 2020-02-24
inotify-tools: upgrade 3.20.1 -> 3.20.2.2

texierp:
Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>


texierp:
in master now

PR #192, opened 2020-02-25
ser2net/gensio

texierp:


kraj:
gensio fails to configure 

http://errors.yoctoproject.org/Errors/Details/392877/

texierp:
> gensio fails to configure
> 
> http://errors.yoctoproject.org/Errors/Details/392877/

Thanks @kraj , I will check that

texierp:
@kraj, PR updated, it seems the problem comes from openssl if libssl is already installed on the Host machine.

Should be fine now.

kraj:
they are in master now

PR #197, opened 2020-03-02
spdlog_1.3.1 allow empty for header only libs

IniterWorker:
By default, BitBake does not produce empty packages.
- opkg_prepare_url_for_install: Couldn't find anything to satisfy 'spdlog'

### DIFF
--- a/meta-oe/recipes-support/spdlog/spdlog_1.3.1.bb
+++ b/meta-oe/recipes-support/spdlog/spdlog_1.3.1.bb
@@ -17,3 +17,5 @@ inherit cmake
 # Header-only library
 RDEPENDS_${PN}-dev = ""
 RRECOMMENDS_${PN}-dbg = "${PN}-dev (= ${EXTENDPKGV})"
+
+ALLOW_EMPTY_${PN} = "1"


kraj:
Please rebase it on top of master branch first. Then backport into warrior

shr-project:
Installing completely empty package doesn't really fix anything on target. Just fix whatever is depending on this empty package.

IniterWorker:
@shr-project, so, what's the best approach?
@kraj, okay, I will do it. Thank you.

shr-project:
Just fix whatever is depending on this empty package.

IniterWorker:
@shr-project, I'm a kind of perplex about your answer.  "Just fix whatever is depending on this empty package." 

You need this package in dev level to package it in an SDK populate. So, in that way, we need to package it, right?

I think we can allow_empty only for the dev-level. So are you okay with it? Or can you give me a hint?

shr-project:
your commit message doesn't show what failed to install with "opkg_prepare_url_for_install: Couldn't find anything to satisfy 'spdlog'" if it was spdlog-dev package, then just set
RDEPENDS_${PN}-dev = ""
in the recipe, like other recipes with empty PN do. If it's something else you're trying to install depending on spdlog, then you need to resolve it there, installing empty package just because some other recipe thinks it depends on it is wrong.

IniterWorker:
I'm in the development stage. I added a new library to update my SDK via my image package list. So, I don't understand why you want I fix other recipes that I didn't depend on it. I want to populate my SDK with a header only in dev-level as well as a standard lib install lib-header into the populated SDK.

As I have said maybe the ALLOW_EMPTY_${PN}-dev = "1" was a good way.

Maybe I missed something for the best way to install a lib header only in the populated SDK.

```
ERROR: placeholder-dev-image-1.0-r0 do_rootfs: Unable to install packages. Command '/workdir/build/tmp/work/varsomam43-poky-linux-gnueabi/placeholder-dev-image/1.0-r0/recipe-sysroot-native/usr/bin/opkg --volatile-cache -f /workdir/build/tmp/work/varsomam43-poky-linux-gnueabi/placeholder-dev-image/1.0-r0/opkg.conf -t /workdir/build/tmp/work/varsomam43-poky-linux-gnueabi/placeholder-dev-image/1.0-r0/temp/ipktemp/ -o /workdir/build/tmp/work/varsomam43-poky-linux-gnueabi/placeholder-dev-image/1.0-r0/rootfs  --force_postinstall --prefer-arch-to-version   install alsa-utils avahi-autoipd avahi-daemon bash bzip2 canutils crda curl devmem2 dosfstools e2fsprogs-mke2fs ethtool findutils fio gdb gdbserver grep gzip i2c-tools ifupdown init-ifupdown iperf3 iproute2 iptables kernel-modules less libcxxopts libdrm-omap1 libpcap1 libserial1 lsof memtester mmc-utils mosquitto mtd-utils nand-recovery nano net-tools netcat-openbsd nmap ntp ntp-tickadj openssh openssh-keygen openssh-sftp-server openssl openssl-bin opkg packagegroup-core-boot parted pciutils procps rsync run-postinsts serial-reader spdlog sysfsutils tcpdump tree udev-extraconf unzip usbutils util-linux util-linux-blkid watchdog wget xz zip' returned 255:
Collected errors:
 * opkg_prepare_url_for_install: Couldn't find anything to satisfy 'spdlog'.

ERROR: placeholder-dev-image-1.0-r0 do_rootfs: 
ERROR: placeholder-dev-image-1.0-r0 do_rootfs: Function failed: do_rootfs
ERROR: Logfile of failure stored in: /workdir/build/tmp/work/varsomam43-poky-linux-gnueabi/placeholder-dev-image/1.0-r0/temp/log.do_rootfs.17881
ERROR: Task (/workdir/meta-placeholder/images/placeholder-dev-image.bb:do_rootfs) failed with exit code '1'
```

shr-project:
ALLOW_EMPTY_${PN}-dev = "1"
won't do anything as ${PN}-dev isn't empty, is it?

Does placeholder-dev-image have spdlog in IMAGE_INSTALL? Then you should fix placeholder-dev-image to IMAGE_INSTALL spdlog-dev not spdlog.

IniterWorker:
My bad, you have right about the non-empty package-split of dev-level of the lib fact. We don't be able to populate the SDK with only adding '-dev'.

For the spdlog-dev, I'm wondering if it's really an excellent way to do it. 
- The question seems pretty tricky.

1) If we decide not to allow the ALLOW_EMPTY, we need to configure a different image between dev (to produce the SDK rootfs) and production.

2) if we decide to allow the ALLOW_EMPTY, we can deal with only one image for dev and prod, but we add an empty package into the rootfs.

Both ways make sense.

Thank you for your arguments,

B. r

graugans:
I had issues with adding spdlog to my SDK as-well. Due to the note in the  commit message of the initial commit when ``spdlog`` was added the hint is to use ``spdlog-dev`` instead of ``spdlog`` what I did.

```
TOOLCHAIN_TARGET_TASK += " \
        spdlog-dev \
        libstdc++-staticdev \
"
```
This solved my issue. Maybe it does help somebody else as-well.

IniterWorker:
@graugans, it's a spotless way to do it.

> TOOLCHAIN_TARGET_TASK: Lists packages that make up the target part of the SDK (i.e. the part built for the target hardware).

> This variable lists packages the OpenEmbedded build system uses when it creates the target part of an SDK (i.e. the part built for the target hardware), which includes libraries and headers.

I'm wondering why packages are not installed into the SDK (ex. spdlog) if they are available only with the suffix "-dev" (ex. spdlog-dev). 

But, whatever the fact, it seems a philosophical matter.
- Some people describe a lib recipe like a regular package manager and install header-only (like Makefile based one)
- Some people describe a lib recipe like a minimal package, with minimal resource conception.

My guts tell me that the minimal way is the good one. But, the principal issue is the fact of both shares in the same place and can consume some extra cost of time to figure out what's the way used by this one.

Thank you for sharing this hint,

B. r

PR #199, opened 2020-03-04
fbgrab: upgrade 1.3 -> 1.3.1

texierp:
Switch SRC_URI to github because 'http://fbgrab.monells.se' is no more reachable
it redirects to a github page.

See full changelog https://github.com/GunnarMonell/fbgrab/releases

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>


texierp:
Now on master

PR #216, opened 2020-04-28
python3-distro: native and nativesdk support

geissonator:
Have a project that needs the native version of this recipe

Signed-off-by: Andrew Geissler <geissonator@yahoo.com>


kraj:
applied via 79dd17c7b05c0951ee791c2f790bf9b83194ef6e

PR #217, opened 2020-04-29
mpv: Remove X11 dependency

parheliamm:
During enable mpv which is based on wayland.
The mpv reqired x11 by default even if mpv support wayland.
Change the depend rules and chang the x11 dectection policy.
After this change, mpv could be build and run successfully based
on wayland

Change-Id: Ibc1a7ce760c0d4dea8069ccaa38fce865a160df7


kraj:
applied via 17fd382f3467e20308924499b7531ae8b789f056

PR #220, opened 2020-05-04
cockpit: fix metainfo.xml file ownership

eroullit:
The `tar -cf - | tar -xf` combo applies an invalid ownership.
This is corrected by patching the install target to use
the `--no-same-owner` `tar` parameter.

Signed-off-by: Emmanuel Roullit <emmanuel.roullit@gmail.com>


hoinmic:
@eroullit That's great! Will you also bring the patch upstream at the cockpit repository?

eroullit:
I have double checked a regular install from source on centos 8:
```
[archie@d47f818d4c65 cockpit-218]$ sudo make install -j 4
[...]
[archie@d47f818d4c65 cockpit-218]$ ls -al /usr/local/share/cockpit/users                                                           
total 224                                                                                                                        
drwxr-xr-x  2 root   root     4096 May  5 19:04 .                                                                                  
drwxr-xr-x 23 root   root     4096 May  5 19:04 ..                                                                               
-rw-r--r--  1 root   root     3222 May  5 19:04 index.html.gz                                                                     
-rw-rw-r--  1 archie archie    751 May  5 18:58 manifest.json                                                            
-rw-r--r--  1 root   root     2605 May  5 19:04 po.ca.js.gz     
[...]
```
`manifest.json` should be owned by `root`, yet it owned by `archie`.
So yes, this patch should be sent to upstream too.

kraj:
fixed with eaf4a9d8a7

PR #227, opened 2020-05-10
Cockpit recipe: Added missing dep on udisks2 for package cockpit-storaged

jsolla:
Cockpit uses udisks2 in order to manage storage on the host, without it
cockpit will just display an error when the storage tab is selected.


kraj:
merged with 353b4d2f98

PR #234, opened 2020-05-20
cockpit: rt-deps for storaged

hoinmic:
No general depdependency on udisks2 (polkit)

cc @jsolla #227


kraj:
applied with 310d99d978

PR #250, opened 2020-07-21
python3-bitstruct: Created recipe

colin-pm:
Created python3 recipe for bitstruct.


akuster:
We are moving away from .inc files now that python2 in its own layers.
Please squish the two into on bb file.

PR #251, opened 2020-07-21
python-cantools: Added recipe

colin-pm:
Added python3 recipe for cantools pypi package.


akuster:
We are moving away from .inc files now that python2 in its own layers.
Please squish the two into on bb file.

PR #256, opened 2020-07-24
libcamera: fix packaging and installation

andrey-konovalov:
libcamera checks if RPATH or RUNPATH dynamic tag is present in
libcamera.so. If it does, it assumes that libcamera binaries are
run directly from the build directory without installing them, and
tries to use resorces like IPA modules from the build directory.
Mainline meson strips RPATH/RUNPATH out at install time (for
meson versions up to 0.54; the things are somewhat changed in 0.55).
But openembedded-core patches meson to disable RPATH/RUNPATH removal.
That's why we need to remove this tag manually in do_install_append().

IPA module is signed (with openssl dgst) after it is built. But
during packaging the OE build system 1) splits out debugging info,
and 2) strips the binaries. So the IPA module *.so file installed
isn't the one which the signature was calculated against. Then
the signature check fails, and libcamera tries to run the IPA
module isolated (in a sandbox), which doesn't work if the IPA
module wasn't designed to run isolated. The easiest way to fix that
is to disable splitting out debug information and stripping the binaries
during packaging with INHIBIT_PACKAGE_DEBUG_SPLIT and
INHIBIT_PACKAGE_STRIP.

Signed-off-by: Andrey Konovalov <andrey.konovalov@linaro.org>

### DIFF
--- a/meta-multimedia/recipes-multimedia/libcamera/libcamera.bb
+++ b/meta-multimedia/recipes-multimedia/libcamera/libcamera.bb
@@ -18,13 +18,20 @@ PV = "202006+git${SRCPV}"
 
 S = "${WORKDIR}/git"
 
-DEPENDS = "python3-pyyaml-native udev gnutls boost"
+DEPENDS = "python3-pyyaml-native udev gnutls boost chrpath-native"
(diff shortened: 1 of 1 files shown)


andrey-konovalov:
This should better be done via the mailing list

PR #264, opened 2020-08-01
Set monit files in temporary locations

shlomow:
Normally monit create its temporary files in $HOME directory.
In case the rootfs is mounted as read-only, monit will not operate.
This patch sets monit configuration to store those files in /var/run

Signed-off-by: Shlomi Vaknin <shlomi.39sd@gmail.com>


kraj:
applied with f7c6adcdef

PR #275, opened 2020-09-11
sdbus-c++: upgrade 0.8.1 -> 0.8.3

sangelovic:
See full changelog at https://github.com/Kistler-Group/sdbus-cpp/blob/v0.8.3/ChangeLog

Signed-off-by: Stanislav Angelovic stanislav.angelovic@protonmail.com


kraj:
Fixed in b6213e4dd5

PR #284, opened 2020-10-06
python3-paramiko: fixing runtime dependencies

simingy:
added rdepends to bcrypt, cryptography, pynacl as required by the
actual python package

Signed-off-by: Siming Yuan <siyuan@cisco.com>


kraj:
this is now in with a952d21168

PR #287, opened 2020-10-19
Add missing dependencies for rsnapshot.

gipi:
From a build of ``core-image-full-cmdline`` with ``rsnapshot`` installed, two dependencies were missing.

I was expecting the command to work out-of-the-box, if there is something that I'm missing let me know.


kraj:
I have pull this into `master-next` as it is needed on master as well. then it can be backported to zeus and dunfell @akuster

kraj:
its in master now see dfd5685ca7

akuster:
thanks. is in stable/dunfell-nut

kraj:
applied with e944a811d39040e9fa453b08fc572aafe1e24ccb

PR #289, opened 2020-11-02
upgrade and de-blacklist nanopb recipe

Jasper-Ben:
The current nanopb recipe is blacklisted due to python2 dependencies.
However, newer releases of nanopb use python3 by default.

This commit updates the used nanopb version to 0.4.3 and removes the
recipe from the blacklist.


kraj:
applied with 24a8d440c5

PR #291, opened 2020-11-17
Thingsboard gw

hmenn:


shr-project:
python3-jsonpath-rw dependency was added to meta-python, but meta-oe
doesn't depend on meta-python, so either python3-jsonpath-rw needs to be
moved to meta-oe (if it doesn't depend on other stuff from meta-python) or
thingsboard-gateway needs to be moved to meta-python.

hmenn:
@shr-project Hi, this commits were added into master-next as it's. Moving thingsboard-gateway under meta-python might not be good. By the way, thingsboard-gateway also has other dependecies from meta-python. So, I think it should be integrated like this. If some one wants to use this recipe, should add meta-python into bblayers as well.

kraj:
> @shr-project Hi, this commits were added into master-next as it's. Moving thingsboard-gateway under meta-python might not be good. By the way, thingsboard-gateway also has other dependecies from meta-python. So, I think it should be integrated like this. If some one wants to use this recipe, should add meta-python into bblayers as well.

@hmenn this changeset caused a bit of work for me. Please look at the commits in meta-oe how its handled with dynamic-layers. I should have noticed this earlier but I test meta-oe and meta-python together and such issues go unnoticed. But @shr-project is right, we do not want the layer to have hard dependency introduced.

kraj:
This has been merged via 
5f17ce66d8
4fb0fb993b
121c94e3c3

hmenn:
@kraj Thanks a lot. I'll check commits carefully.

PR #66, opened 2018-02-07
devtools/php: remove the failing ${D}/${TMPDIR} code

MaxKellermann:
Appending ${TMPDIR} to ${D} doesn't make any sense, because both are
absolute paths.  And additionally, the code fails:

 rmdir: failed to remove '/usr/src/oe/tmp-musl/work/core2-64-oe-linux-musl/php/7.1.9-r0/image//usr': Directory not empty

PR #292, opened 2020-11-20
[meta-oe] devtools/php: remove the failing ${D}/${TMPDIR} code

MaxKellermann:
Appending ${TMPDIR} to ${D} doesn't make any sense, because both are
absolute paths.  And additionally, the code fails:

 rmdir: failed to remove '/usr/src/oe/tmp-musl/work/core2-64-oe-linux-musl/php/7.1.9-r0/image//usr': Directory not empty

Signed-off-by: Max Kellermann <max.kellermann@gmail.com>

(This is the successor of #66 which I cannot update because I had deleted my fork long ago. @kraj)

### DIFF
--- a/meta-oe/recipes-devtools/php/php_7.4.9.bb
+++ b/meta-oe/recipes-devtools/php/php_7.4.9.bb
@@ -155,7 +155,6 @@ do_install_prepend_class-target() {
 # fixme
 do_install_append_class-target() {
     install -d ${D}${sysconfdir}/
-    rm -rf ${D}/${TMPDIR}
     rm -rf ${D}/.registry
     rm -rf ${D}/.channels
     rm -rf ${D}/.[a-z]*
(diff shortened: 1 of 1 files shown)


shr-project:
I guess you're seeing this issue because our TMPDIR starts with /usr while for most people it will be something line /home/foo or /OE etc which don't conflict with valid files in ${D}/usr

But it made sense in 2013 when it was introduced https://git.openembedded.org/meta-openembedded/commit/?id=357915b6d11dfa612fbadaa4966a0b3e0f332467 and if php still does install files in this wrong location, then it still needs to be fixed not just reverted.

MaxKellermann:
If PHP installs files in those insane locations, then this bug should be patched in the PHP build system, instead of trying to fix it up after the fact.

shr-project:
> If PHP installs files in those insane locations, then this bug should be patched in the PHP build system, instead of trying to fix it up after the fact.

agreed, it was just work around in 2013 and nobody sent patch to properly fix that since then

kraj:
@MaxKellermann thanks, has this been tested in non /usr tmpdir env?

kraj:
this has been merged via 5cd29d53b349b566e2f05c46c5da56b382b95465

PR #298, opened 2020-12-07
[master] Fix unreachable SRC_URIs referring to the Debian pool

rohieb:
Originally I just wanted to fix up openct, which is no longer available in the Debian pool, but the other two were also discovered while trying to fix it.


rohieb:
Mmmh, the openct tarballs from github are apparently different from the other ones, and the patches no longer apply… This needs some more work.

rohieb:
Alright, works for me now.

kraj:
this is now merged into master

rohieb:
@akuster At least the first commit (openct) is also an issue on dunfell. Do you want a separate PR, or can you cherry-pick?

PR #299, opened 2020-12-13
[master][dunfell] openct: cleanup and allow building as native package

rohieb:
* Allow building openct as native package
* Clean up `do_install`:
  - Use `sysconfdir` and `nonarch_libdir` instead of hard-coded paths, and be
    consistent with the slashes after `${D}`
  - Install udev rules into `/usr/lib/udev` instead of `/etc/udev` so they can
    be overwritten in `/etc`, which has higher priority. Also use `/usr/lib`
    instead of `/lib`, latter of which isn't searched by udev (see
    [manpage][1]).
  - Don't create `/usr/lib/udev`, it is already created by `install -D` later
    with the default umask, which is fine.
  - Make use of `do_install[cleandirs]`.
* Remove lines that first created `/var/run/openct` and then removed it again

This is for master, but applies to dunfell as well.

### DIFF
--- a/meta-oe/recipes-support/openct/openct_0.6.20.bb
+++ b/meta-oe/recipes-support/openct/openct_0.6.20.bb
@@ -46,7 +46,6 @@ FILES_${PN} += " \
     ${nonarch_base_libdir}/udev \
     ${libdir}/openct-ifd.so \
     ${libdir}/pcsc \
-    /run/openct/status \
 "
 
 FILES_${PN}-dbg += " \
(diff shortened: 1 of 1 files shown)


meta-oe/recipes-support/openct/openct_0.6.20.bb
```diff
@@ -56,33 +55,29 @@ FILES_${PN}-dbg += " \
...
-
-    install -dm 755 ${D}${localstatedir}/run/openct
-    touch ${D}${localstatedir}/run/openct/status
-    chmod 644 ${D}${localstatedir}/run/openct/status
-
-    rm -r ${D}/${localstatedir}/run
+    install -Dpm 644 etc/reader.conf ${D}${sysconfdir}/reader.conf.d/openct.conf
 }
+
+BBCLASSEXTEND="native"
```
kraj, line 83: usually, a space on both sides of `=` is norm.

kraj:
the changes are in master now, I tweaked it myself to address the comment.

rohieb:
Thank you!

PR #307, opened 2021-01-14
python3-sh: remove python3-tests from RDEPENDS

shlomow:
python3-tests installs the unit tests of python
as well as stuff that no one should care about
in its rootfs. Thus, it is not a runtime dependency
of python3-sh.

Signed-off-by: Shlomi Vaknin <shlomi.39sd@gmail.com>


kraj:
its in master now with 2ed25eb4ef

PR #318, opened 2021-03-04
python3-gpsd-py3: Added recipe

colin-pm:
Create Python3 recipe for gpsd-py3.

### DIFF
--- a/meta-python/recipes-devtools/python/python3-gpsd-py3_0.3.0.bb
+++ b/meta-python/recipes-devtools/python/python3-gpsd-py3_0.3.0.bb
@@ -1,0 +1,16 @@
+DESCRIPTION = "This is a library for polling gpsd in Python3"
+HOMEPAGE = "https://github.com/MartijnBraam/gpsd-py3"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://setup.py;beginline=10;endline=10;md5=c2d9994c57f0444e39f1dab19af50254"
+SRC_URI[md5sum] = "041ce56e8879e2104b4d54c8119cd529"
+SRC_URI[sha256sum] = "2908d3bd78dfb6720ecfe22f97e139b5a4a198f38df3a77215cf644a33513192"
+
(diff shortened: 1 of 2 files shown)


meta-python/recipes-devtools/python/python3-gpsd-py3_0.3.0.bb
```diff
@@ -0,0 +1,17 @@
...
+SRC_URI[sha256sum] = "2908d3bd78dfb6720ecfe22f97e139b5a4a198f38df3a77215cf644a33513192"
+
+PYPI_PACKAGE = "gpsd-py3"
+
+inherit pypi setuptools3
+
+CLEANBROKEN = "1"
+
+RDEPENDS_${PN} += " \
+    gpsd \
```
kraj, line 15: gpsd comes from meta-oe layer, but its not present in layer dependencies. for meta-python, so this should be added in layer.conf

kraj: Also add it to packagegroup-meta-python since its a new recipe

kraj (changes requested):

meta-python/recipes-devtools/python/python3-gpsd-py3_0.3.0.bb
colin-pm: On second thought, I don't think gpsd is a runtime dependency.  I forgot that you can connect to a remote gpsd service, so I will remove gpsd from the runtime dependencies.

I'll add the package to packagegroup-meta-python.  To confirm, that's in meta-python/recipes-core/packagegroups/packagegroup-meta-python.bb, correct?

kraj: right thats the packagegroup recipe to add to

colin-pm:
gpsd has been removed from runtime dependencies and I added python3-gpsd-py3 to packagegroup-meta-python

kraj:
this should be fixed now with 9991671eb3

PR #320, opened 2021-03-10
libbpf: use pkg-config

teknoraver:
The libbpf makefile uses pkg-config to get the libelf build flags and
file paths.
Inherit pkgconfig so the install target can copy the binaries in the
sysroot, which are needed by other packages, like bcc.

While at it, remove "native" which is unneeded for this recipe.

Signed-off-by: Matteo Croce <mcroce@microsoft.com>


teknoraver:
Hi @kraj,

this is needed for kraj/meta-clang#441 to work, can you please have a look?

kraj:
this is merged with ac05388b628aa2d7bec996438eb786b2cd57f451

PR #322, opened 2021-03-23
guider: Upgrade 3.9.7 -> 3.9.8

iipeace:
Upgrade to release 3.9.8


iipeace:
@kraj ,
please review this PR :)

kraj:
@iipeace these pulls are confusing since there is rebasing happening constantly after pull is published, not only rebasing but the SRCREV in recipe is changing too. I ended up picking a version that has `SRCREV = "bab995ce718b58d4b5fee594771884dea238d70a"` and then this revision disappeared as reported by @shr-project on mailing lists [1]

My humble request is to finalize your side of updates before publishing a pull for OE, so such things can be avoided and we don't have to waste time in fruitless iterations.

[1] https://lists.openembedded.org/g/openembedded-devel/message/90386

iipeace:
@kraj , @shr-project ,
I am really sorry for your jobs caused by my force pushes.
I have not known your works outside GitHub.
I had updated commits to fix new bugs because I thought you didn't checked this PR yet.
Next time, I will send a new PR after fixing a new version and adding a new tag.
Thanks.

kraj:
@iipeace OE workflow is primarily email/mailing-list based so we don't merge the pulls directly ( yet ) but we do accept them where we pickup the pulls and then push them manually into `master-next`, this is a bit crude but that's what we have. So if you add a comment before rebasing that will at least inform maintainers that you are working on a rebase or something

PR #338, opened 2021-04-13
Safe PEP8

persianpros:


kraj:
you should be adding `Signed-off-by: ` to every commit see https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

kraj:
this is now merged

PR #340, opened 2021-04-16
strongswan: Make PACKAGECONFIG a default value

jhershbe:
Change from a weak default to a default in the definition of the PACKAGECONFIG.

In https://github.com/flihp/meta-measured/blob/master/networking-layer/recipes-support/strongswan/strongswan_5.%25.bbappend the PACKAGECONFIG is appended to, so if the definition is weak here, the variable will be empty when the bbappend attempts to add to it.


kraj:
merged with 90c04acced

Issue #327, opened 2021-03-26
UFW in dunfell: 'python': No such file or directory


jostor:
When running UFW 0.33, built from the dunfell branch, I get this error:` /usr/bin/env: 'python': No such file or directory.` I guess it is because python2 is removed from the dunfell branch. But shouldnt it be possible to use UFW with python3? And shouldn't that be default on the dunfell branch?

shr-project:
Using python3 with dunfell would be definitely preferred. Patches are always welcome.

akuster:
This should fail on master as well. There have not been any changes to the recipe zeus

Silcet:
I actually have a patch for this. Let me clean it up and I'll link the PR here

Silcet:
I have fixed this issue for my system. I'll create a pull request soon

PR #344, opened 2021-04-23
ufw: fix python shebang

Silcet:
[meta-openembedded ticket #327] --
https://github.com/openembedded/meta-openembedded/issues/327

The python version in the shebang at the begining of the ufw script
should be the same one as the version the setup.py script was called
with.

The fix in patch "setup-only-make-one-reference-to-env.patch"
depends on sys.executable returning "/usr/bin/env pythonX". However,
it returns "/usr/bin/pythonX". Using sys.version_info we can get the
major version of the python used to called the script and append
that to the shebang line so it works as intended.

Signed-off-by: Silcet <camorga1@gmail.com>

### DIFF
--- a/meta-networking/recipes-connectivity/ufw/ufw/fix-dynamic-update-of-python-shebang.patch
+++ b/meta-networking/recipes-connectivity/ufw/ufw/fix-dynamic-update-of-python-shebang.patch
@@ -1,0 +1,57 @@
+From b961a7fceb5654c283c3f987bee593d52abaf1f5 Mon Sep 17 00:00:00 2001
+From: Silcet <camorga1@gmail.com>
+Date: Mon, 26 Apr 2021 07:47:02 +0000
+Subject: [PATCH] ufw: Fix dynamic update of python shebang
+
+[meta-openembedded ticket #327] -- https://github.com/openembedded/meta-openembedded/issues/327
+
+The python version in the shebang at the begining of the ufw script
+should be the same one as the version the setup.py script was called
+with.
+
+The fix in patch "setup-only-make-one-reference-to-env.patch"
+depends on sys.executable returning "/usr/bin/env pythonX". However,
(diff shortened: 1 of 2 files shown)


kraj (changes requested):
this needs to be added to `SRC_URI` as well :)

Silcet:
I'm a rookie so I made a rookie mistake :) . Fixed now.

Silcet:
Also, will this be backported or should I make a new PR for dunfell?

kraj:
once merged into master, backport that into dunfell and create a new pull

kraj (commented):
squash it into single commit.

Silcet:
Done. Sorry for the mistakes. First time contributing to a big project :)

kraj:
Fails to build see

https://autobuilder.yoctoproject.org/typhoon/#/builders/88/builds/1170

kraj (changes requested):
this does not compile as reported

Silcet:
Working on the fix atm. I'll push the changes in a bit

Silcet:
@kraj I've been working more with UFW these days and although it builds fine and has the proper python3 shebang to work now it has many other flaws. The configuration files are not read properly and other issues (I can elaborate if you are interested). I'm testing with a bumped version to 0.36 which seems to work much better. Would it make sense to just drop this PR and make another one with a working 0.36 instead?

kraj:
> @kraj I've been working more with UFW these days and although it builds fine and has the proper python3 shebang to work now it has many other flaws. The setup.py modifies source files in a way that is incompatible with how distutils is used in openembedded (I can elaborate if you are interested). I'm testing with a bumped version to 0.36 which seems to work much better. Would it make sense to just drop this PR and make another one with a working 0.36 instead?

that would be fine too

kraj:
applied with 5194af1afd

PR #351, opened 2021-05-03
ufw: bump version to 0.36 and add services

Silcet:
This commit brings the version of ufw up to 0.36 since version 0.33
had some problems:
* The setup.py calls sed to replace some variables in the source
code with the correct paths. However, this is done using a hardcoded
path and conflicts with distutils
* The python shebang was not properly corrected in setup.py, leading
to a script that only run if there is a python symlink to python2 or
python3

The first issue is addressed by the bump in version, while the second
one is fixed in patch 0003 of the recipe.

Also, the new version provides examples for systemd service and
sysvinit scripts to autostart ufw. These are added into the recipe
now.

Signed-off-by: Silcet <camorga1@gmail.com>


kraj:
merged with 0a0894b78e

PR #359, opened 2021-05-17
[master][meta-oe] libiio: add serial_backend support plus minor tweak

nunojsa:
The first patch in the series adds support for the SERIAL_BACKEND. With [this change](https://github.com/analogdevicesinc/libiio/pull/680) in libiio, the backend options will be enabled by default and then dependencies are looked for. Hence, this would make the cmake configuration to fail as libserialport would not be present. Before, things worked because the libiio build system was only presenting cmake options if the dependencies were found in the host system. With this patch, things should automatically work as soon as we bump libiio version.
The second patch is just a minor tweak as libxml2 should always be treated as DEPENDS.


nunojsa:
Hi @kraj, Would it be acceptable to backport this to dunfell? I'm not aware of the policy here and if only fixes are "good" to be backported...

kraj:
I think it is backportable. I will let @akuster decide though

nunojsa:
Thanks... If needed, I can open a new PR targeting dunfell

PR #369, opened 2021-06-04
linuxptp: improvements

pberginkonsult:
A couple of improvements to the linuxptp recipe. We use the install function from the makefile in order to install all produced binaries and manual pages. Also ship and package example configurations in a separate package. Inspiration is taken from how debian package linuxptp.

### DIFF
--- a/meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
+++ b/meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
@@ -1,4 +1,5 @@
 DESCRIPTION = "Precision Time Protocol (PTP) according to IEEE standard 1588 for Linux"
+HOMEPAGE = "http://linuxptp.sourceforge.net/"
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://COPYING;md5=b234ee4d69f5fce4486a80fdaf4a4263"
 
--- a/meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
+++ b/meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
@@ -10,14 +10,12 @@ SRC_URI = "http://sourceforge.net/projects/linuxptp/files/v${PV}/linuxptp-${PV}.
 SRC_URI[md5sum] = "2264cb69c9af947028835c12c89a7572"
 SRC_URI[sha256sum] = "f58f5b11cf14dc7c4f7c9efdfb27190e43d02cf20c3525f6639edac10528ce7d"
 
-EXTRA_OEMAKE = "ARCH=${TARGET_ARCH} EXTRA_CFLAGS='${CFLAGS}'"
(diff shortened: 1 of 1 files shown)


meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
```diff
@@ -10,14 +10,12 @@ SRC_URI = "http://sourceforge.net/projects/linuxptp/files/v${PV}/linuxptp-${PV}.
 SRC_URI[md5sum] = "2264cb69c9af947028835c12c89a7572"
 SRC_URI[sha256sum] = "f58f5b11cf14dc7c4f7c9efdfb27190e43d02cf20c3525f6639edac10528ce7d"
 
-EXTRA_OEMAKE = "ARCH=${TARGET_ARCH} EXTRA_CFLAGS='${CFLAGS}'"
+EXTRA_OEMAKE = "ARCH=${TARGET_ARCH} EXTRA_CFLAGS='${CFLAGS}' mandir=${mandir}"
```
kraj, line 14: should this `inherit manpages` perhaps ?

meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
```diff
@@ -17,6 +17,13 @@ export KBUILD_OUTPUT="${RECIPE_SYSROOT}"
 
 do_install() {
     oe_runmake install DESTDIR=${D} prefix=${prefix}
+
+    # Install example configs from source tree
+    install -d ${D}${docdir}/${PN}
+    cp -r ${S}/configs ${D}${docdir}/${PN}
```
kraj, line 23: I wonder if using something like below should be used
```
cp -R --no-dereference --preserve=mode,links
```

kraj (changes requested):

meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
pberginkonsult: The man pages in linuxptp is not built, they are in the repo from the beginning. There is neither a configuration option to enable/disable the installation of them. (https://salsa.debian.org/multimedia-team/linuxptp/-/blob/master/makefile#L86) For that reason I don't see any need for 'inherit manpages' that create the PACKAGECONFIG manpages.

Then I see in manpages.bbclass that there is some indexing in pkg_postinst (http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/manpages.bbclass#n15) that probably coudl be useful for linuxptp.

So I'm a bit unsure, please advise if you have an opinion. We can probably make the assumption that inherit manpages does not harm anything.

meta-oe/recipes-connectivity/linuxptp/linuxptp_3.1.bb
pberginkonsult: I have looked in oe-core and meta-openembedded and can not find a common pattern around this. But agree that your suggestion is better if in the future links show up among the installed files.

kraj:
thanks I have staged it in `master-next`

kraj (approved):
LGTM

kraj:
this is in master now.

PR #393, opened 2021-07-13
Meta python merge

threexc:


crazystick:
@threexc the recipe for pyusb should include libusb1 as RDEPENDS in addition, as the automatic runtime dependency calculation doesn't seem to work for python libraries

kraj:
> @threexc the recipe for pyusb should include libusb1 as RDEPENDS in addition, as the automatic runtime dependency calculation doesn't seem to work for python libraries
> 
> If you want this as a separate issue, let me know

I think it will be good to fix this. This patch is proposed on oe devel mailing list. It will be good to respond to submitter to add it and we can hold this patch

kraj:
merged

PR #398, opened 2021-07-26
htop: Add ncurses-terminfo-base to RDEPENDS

ptsneves:
Without it there are no terminal configurations on the target
and htop refuses to run.

Simple patch :)


kraj:
in master with b5d74f8a6b

Issue #147, opened 2019-10-11
Is adbd supposed to work with normal linux kernel? I tried and failed.


sprhawk:
I'm using warrior branch ( 2.7.1 ), and put

```
IMAGE_INSTALL_append = " android-tools"
```

to install adb daemon, but it doesn't start up, and I see /usr/bin/android-setup-gadget, there is no `/system/build.prop ` , `/sys/class/android_usb/*` files to operate on

`adbd --help` will return segmentation fault

kraj:
this should be a bug, please post any stack traces or debug information that you can collect

sprhawk:
@kraj I want to know: is it supposed to work on a normal linux build? I know there are multiple partitions on android's system partition, but there are only two ( boot and rootfs ) partitions on normal linux build. Is this the reason adbd cannot work?

But if adbd is supposed to work on a normal linux, I would like to pay more time on finding the reason.

Thanks

kraj:
@shr-project might have more to add on adb usage on OE, but I think the fact that its crashing is a bug that should be fixed

shr-project:
It's not supposed to work on a normal linux build.

We (LuneOS) are using android-tools-native to create fastboot images and target android-tools together with Halium which provides necessary support in kernel and rootfs, see meta-android layer for example how to use it.

kraj:
@shr-project does it make sense then to limit the recipe to built native and nativesdk alone and not for target

shr-project:
We're using target version as well as mentioned above, limiting it to native(sdk) would force us to add bbappend re-enabling it for target.

tewarid:
I am now able to run adb daemon with systemd on a linux device (rpi zero w) with the following in `local.conf`

```bitbake
IMAGE_INSTALL_append = " android-tools-adbd"
PREFERRED_PROVIDER_android-tools-conf = "android-tools-conf-configfs"
```

You'll need to `touch /var/usb-debugging-enabled` to enable usb debugging, as that is a precondition for systemd to start the service.

I am connecting the rpi zero w to host using USB OTG port, and meta-raspberrypi BSP layer requires the following in `local.conf` to enable USB gadget/peripheral mode

```bitbake
ENABLE_DWC2_PERIPHERAL = "1"
```

sprhawk:
cool

sprhawk:
> I am now able to run adb daemon with systemd on a linux device (rpi zero w) with the following in `local.conf`
> 
> ```
[...]

PR #399, opened 2021-07-26
Add flag to enable adbd service (#147)

tewarid:
android-tools-adbd service can be enabled in the image using  `USB_DEBUGGING_ENABLED = "1"` in `local.conf`.


kraj:
I wonder if it should be controlled via do_rootfs post-processing instead, so you can use same package to build with and without usb debugging

tewarid:
I did try it but I haven't been able to get do_rootfs post-processing to work with wic bmap image format used by meta-raspberrypi.

kraj:
Can you turn this into a PACKAGECONFIG that way someone can enable it via a bbappend.

kraj (changes requested):
turn it into packageconfig or rootfs post processing cmd.

tewarid:
I'll close this PR and submit another later based on requested changes.

PR #400, opened 2021-07-26
Suppress eol in functionfs setup scripts (#147)

tewarid:
Stray newline character causes errors in functionfs setup scripts used by android-tools-adbd.service, when using musl libc and/or toybox.


kraj:
applied with db5f487344

PR #403, opened 2021-07-30
Adds a recipe for bmon (and libconfuse)

mandraga:
Buildroot has it and it is usefull to analyse network bandwidth problems. So it could be usefull in oe.


meta-networking/recipes-support/bmon/bmon_2.1.0.bb
```diff
@@ -0,0 +1,13 @@
+DESCRIPTION = "Portable Bandwidth Monitor and rate estimator"
+HOMEPAGE = "http://people.suug.ch/~tgr/bmon/"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+# only works with libnl-0.5.0
+DEPENDS = "libnl libconfuse ncurses"
+
+#SRCREV = "1b3f11bde315e221474f7d066ce4efb4ff4d39e3"
+SRC_URI = "git://github.com/tgraf/bmon.git;branch=master;tag=v4.0"
```
kraj, line 9: Please do not use tags, lets use SRCREV like you have on commented line above, using tags means we need network to build always.

kraj (changes requested):
follow commit guidelines mentioned here to format the commit msg
https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
```diff
@@ -0,0 +1,27 @@
+DESCRIPTION = "libConfuse is a configuration file parser library"
+LICENSE = "CUSTOM"
+LIC_FILES_CHKSUM = "file://LICENCE;md5=42fa47330d4051cd219f7d99d023de3a"
+
+DEPENDS = "gettext-native"
+
+#SRCREV = "1b3f11bde315e221474f7d066ce4efb4ff4d39e3"
+SRC_URI = "git://github.com/libconfuse/libconfuse.git;branch=master;tag=v3.3"
```
kraj, line 8: same issue, do not use tags

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
```diff
@@ -0,0 +1,27 @@
...
+
+do_configure_prepend(){
+    cd ${S}
+    ${S}/autogen.sh
+}
+
+do_compile(){
+    cd ${S}
+    oe_runmake
+}
```
kraj, line 22: you can inherit `autotooks-brokensep` instead of `autotools` if the package is not buildable out of tree. and then you wont need to override `do_compile` and `do_intall`

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
```diff
@@ -0,0 +1,27 @@
...
+
+do_install(){
+    cd ${S}
+    oe_runmake DESTDIR=${D} install
+}
```
kraj, line 27: this also should go away

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
mandraga: Ok, I only kept do_configure, the other worked. Because I use an older version of oe, I did not use autotooks-brokensep but only B = "${S}".

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
mandraga: Done

meta-networking/recipes-support/libconfuse/libconfuse_3.3.bb
mandraga: Yes, I could delete it once I added B = "${S}".

kraj:
thanks for rework. Can you now create fresh pull with two commits one for each recipe ? and also following the commit guildelines ?

mandraga:
Yes I will create something clean.

PR #417, opened 2021-08-20
bmon and libconfuse recipes

mandraga:
This pull request adds the recipe for bmon, a network monitor tool using ncurses. It also adds libconfuse which is a dependency not used elsewhere.
It is a rework of a previous pull request.

By the way I was not confident about submiting a patch using gmail for this. But I see how it can be done, next time.


kraj:
can you address few comments I posted. Its looking good otherwise.

kraj:
merged with 97daeb1c1c e56b987c21 and 7b68cad206

PR #424, opened 2021-08-28
fix Syntax for Hardknott

kueken:


akuster:
thanks for the patch. A patch to fix this was already sent to the mailing list and is now in hardknott-next pending merge

kueken:
ok 
had seen the PR too late.
So I close the PR here.

PR #425, opened 2021-08-28
fix Syntax for Hardknott

kueken:


akuster:
this fix is pending merge to hardknott. A patch to address this was sent earlier in the week.
thanks.

kueken:
Thanks

PR #430, opened 2021-09-06
libbpf: remove kernel configuration dependency

pdmorrow:
This dependency introduces a cycle with pahole:

```
kernel --> pahole --> libbpf
  ^                      /
   \                    /
    `------------------<
```

pahole is a build time dependency of the kernel which is required to
generate BTF symbols. The dependency on the kernel configuration is nice
to have but makes using pahole during kernel compile time impossible.

Furthermore this dependency doesn't make sense when building libbpf
purely to ship it via a SDK.

Signed-off-by: Peter Morrow <pemorrow@linux.microsoft.com>


kraj:
merged with 4c21b66338

PR #431, opened 2021-09-06
cryptsetup: Add runtime dependency on lvm2-udevrules for udev

klausenbusk:
Without the udevrules cryptsetup luksOpen will be hanging with "Udev
cookie 0xd4de0f6 (semid 5) waiting for zero".


kraj:
merged with 9b02aa1220

PR #440, opened 2021-09-20
poco: update to 1.11.0

bachp:
See https://pocoproject.org/docs/99100-ReleaseNotes.html

Signed-off-by: Pascal Bach <pascal.bach@siemens.com>


kraj:
applied with 8527bd0ef8

PR #448, opened 2021-10-05
rapidjson: override hard-coded lib install dir

rohieb:
The default value for `LIB_INSTALL_DIR` is set to `${prefix}/lib` in CMakeLists.txt. However, in multilib builds, the package split picks up the files from `${prefix}/lib64` or `${prefix}/lib32`, which leads to QA issues in do_package about left-over files in `/usr/lib/pkgconfig` and `/usr/lib/cmake`.

The fix applies to hardknott too.


kraj:
applied with 8e6e4e594d

PR #459, opened 2021-10-25
openldap: upgrade 2.4.58 -> 2.5.8

salmanisd:
- dropped retired backends (bdb, hdb, shell)
- back-monitor is now built as part of slapd
- added asyncmeta and wt backends
- dropped patches for functionalities which don't
  exist anymore


kraj:
@salmanisd will you be able to provide update with fix for python3-ldap here ?

salmanisd:
@kraj 
proper fix would be to modify python-ldap to support openldap v2.5 but according to the python-ldap devs, adding this is not trivial. 
Meanwhile there are some workarounds like mentioned here https://github.com/python-ldap/python-ldap/issues/432

I'm not sure how to proceed further with the fix.

kraj:
> @kraj proper fix would be to modify python-ldap to support openldap v2.5 but according to the python-ldap devs, adding this is not trivial. Meanwhile there are some workarounds like mentioned here [python-ldap/python-ldap#432](https://github.com/python-ldap/python-ldap/issues/432)
> 
> I'm not sure how to proceed further with the fix.

we can apply workaround mentioned in https://github.com/python-ldap/python-ldap/issues/432#issuecomment-958084741
for OE, since we control theversion of openldap for OE and we know it will be 2.5+

salmanisd:
works for me... 
I will add a python-ldap change commit in this branch next week

salmanisd:
@kraj 
I have added the python3-ldap fix.

kraj:
> @kraj I have added the python3-ldap fix.

thanks, staged in master-next

PR #465, opened 2021-10-28
Feature/linter

HerrMuellerluedenscheid:
I propose to add a linter. This will decrease review time and make all receipes more homogeneous in terms of style, thus easier to maintain.

[This linter](https://github.com/priv-kweihmann/oelint-adv) is pretty configurable. I think it's starting pretty strict (which I think is good). But in some cases a more flexible tuning rather with info/warning rather than error can be discussed.

For demonstration purposes I touched the `python3-django_2...` recipe. Have a look at the linter output: https://github.com/openembedded/meta-openembedded/runs/4036977361?check_suite_focus=true

### DIFF
--- a/.github/workflows/oe-linter.yml
+++ b/.github/workflows/oe-linter.yml
@@ -37,5 +37,5 @@ jobs:
         run: |
           for file in ${{ steps.changed-files.outputs.all_modified_files }}; do
             echo "spell checking $file"
-            codespell $file
+            codespell $file --skip=oelint-rulefile.json
           done


shr-project:
There used to be automated review script integrated somehow with patchwork and was reporting various issues for patches submitted to oe-core ML.

http://git.yoctoproject.org/cgit/cgit.cgi/patchtest

Here is an random example how the replies from patchtest looked:
https://yhbt.net/lore/all/20181102030313.13842.45534@do.openembedded.org/

@rpurdie mentioned that someone is working on patchwork again, I don't know who, maybe the linter could be integrated there as well.

HerrMuellerluedenscheid:
Hey @all, just wanted to check back in here. @priv-kweihmann and I worked a little to improve the code or `oelint` during the last week. E.g. now there is an option `--relpaths` which only shows relative paths in the output which removes a lot of unnecessary text. Checkout the example of this associated MR: https://github.com/openembedded/meta-openembedded/runs/4106189722?check_suite_focus=true

One more proposition is in the pipeline [in this issue](https://github.com/priv-kweihmann/oelint-adv/issues/269)
Your feedback and input is highy appreciated!

I think it's good to push this issue a little because a proper linter can improve the learning curve a lot when starting to work in a new domain.

HerrMuellerluedenscheid:
Hey,

From my point of view the linter can be merged. It's a starting point, thus there may be some tweaking required in the future. But I think it's good start to get used to it.
Cheers

rpurdie:
It probably falls to me to highlight an issue here. Developing linters is great but doing it outside the established development method of the project potentially creates some problems. I mentioned this work to others and they're unaware of it as it isn't on the mailing lists where people expect this kind of development to be done. As such, many developers haven't an opportunity to comment on it or give feedback. I appreciate some people dislike the mailing lists and want to use github only. There are other developers with the opposite view too. Allowing pull requests against meta-openembedded was an experiment which Khem as the maintainer was wanting to undertake but this pull request does move things into trickier territory. If such a change were merged, it could create a lot of friction as patches are meant to be reviewed on the mailing list and feedback from the linter wouldn't make it there. This would mean that either, often it would be ignored, or patches would end up blocked for unknown reasons and neither outcome would be good. If it effectively mandates github pull requests, that changes the development model and that isn't something which should happen by stealth. I'm well aware of the two sides to all of this and as a maintainer of OE-Core, I know all too well the issues involved so in many ways I don't want to get involved here but I do feel I need to mention it.

priv-kweihmann:
I honestly don't see the issue here - the proposed change is working only on the changed files in a guthub PR, giving almost immediate feedback - so I don't see where a mailing list has to involved, as it simply doesn't affect ML users at all.
And to be totally frank patchwork isn't up to the task either - so for me it boils down to the question if the project does want just a feature or not? (currently I feel the answer is rather not)

HerrMuellerluedenscheid:
@rpurdie I understand your point. But I also don't think this is a real issue. If contributors prefer to send patches to the mailling list to get pure manual reviews they can still to that. If instead they prefer direct and immediate feedback to be sure that they adhere to a proper structure and did not forget any critical parts, they can open a merge request which is still followed by a manual review with the advantage for the reviewer that he/she saves time as basic linting has already been taken care of. 
I personnally share the oppinion that immediate feedback is better, saves time and leads to less frustration as productivity is increased.

Alternatively, if a github action is the wrong place for this linter, could this somehow be integrated into the ML workflow? It's out of my hands then but I think this could be an alternative.

rpurdie:
> I honestly don't see the issue here - the proposed change is working only on the changed files in a guthub PR, giving almost immediate feedback - so I don't see where a mailing list has to involved, as it simply doesn't affect ML users at all. And to be totally frank patchwork isn't up to the task either - so for me it boils down to the question if the project does want just a feature or not? (currently I feel the answer is rather not)

It really depends where this ends up going. If for example you or others "clean up" a recipe to pass the linter checks and some changes then introduce a regression from the linter perspective, is that going to be a problem? If not and it is just extra validation for patches that might be ok but it isn't clear here whether this has been thought about or not.

Also, I know oelint has a reputation of being a bit too pedantic and people don't agree with everything it reports. That leads to a risk that changes here to pass the linter might not be changes the wider community wants.

By discussing this in a github pull request we're already creating a split community where key changes are no longer being discussed in a definitive place where people can watch for and expect them and that in itself is a big issue.

HerrMuellerluedenscheid:
> It really depends where this ends up going.

We will never know if we don't event take the first step. 

> If for example you or others "clean up" a recipe to pass the linter checks and some changes then introduce a regression from the linter perspective, is that going to be a problem? If not and it is just extra validation for patches that might be ok but it isn't clear here whether this has been thought about or not.

I'm not sure I can follow here. Can you elaborate a little? There is no auto-fixing happening. Did you take a look at what the linter does?

> Also, I know oelint has a reputation of being a bit too pedantic and people don't agree with everything it reports. That leads to a risk that changes here to pass the linter might not be changes the wider community wants.

It may have that reputation because people didn't configured the linter. By default it's oppinionated and strict. Those aspects have been discussed and addressed throughout the thread. What is an error or a warning is [highly customizable](https://github.com/openembedded/meta-openembedded/pull/465/files#diff-cb34d740044330e2ccbfa33d507e2bf7a08a3f122775d96f7c172cbd047957a3), as are the format and style of the output. Thus, I invite you to give it a try and provide specific feedback on linter settings to help finding a reasonable configuration. I think your input would be a tremendous help.

> By discussing this in a github pull request we're already creating a split community where key changes are no longer being discussed in a definitive place where people can watch for and expect them and that in itself is a big issue.

I'll post to the mailing list to draw other peoples' attention to this thread.

rpurdie:
> > If for example you or others "clean up" a recipe to pass the linter checks and some changes then introduce a regression from the linter perspective, is that going to be a problem? If not and it is just extra validation for patches that might be ok but it isn't clear here whether this has been thought about or not.
> 
> I'm not sure I can follow here. Can you elaborate a little? There is no auto-fixing happening. Did you take a look at what the linter does?

I'm assuming the idea behind having a linter run on commits is that the commits would be adjusted to make the linter happier. My worry was someone working through a different workflow regressing changes and adding issues back. I can imagine some scenarios where that could upset the people improving the linter warnings. I also know that some people don't like linter only changes to the code and I don't have a clear idea how the linter is planned to be used here so that was also partly what I was referring to. Would someone be going recipe by recipe to lint meta-oe to pass the linter or is this only on newly added lines? I have used oelint before, I haven't looked at exactly how it is integrated here into github.

> It may have that reputation because people didn't configure the linter. By default it's oppinionated and strict. Those aspects have been discussed and addressed throughout the thread.

Keep in mind that you've discussed it with a limited number of people in a place most people don't know to look.

I'm not against the idea of the linter, my biggest worry here is that we're now effectively creating two different workflows for developing features and merging patches. The patch on mailing list vs pull request issue has been discussed before and I'm very worried about fragmenting the project and it's developers. As the OE-Core maintainer, I know an inevitable question will be "when do we enable this for core?" and I also know how that will cause friction elsewhere. I appreciate I'm looking into the future but for better or worse I do need to do that and I do have some idea of what can happen from experience. It does make me look rather negative unfortunately which isn't my intention. Equally, if I just ignore this until it becomes a bigger issue, that isn't fair to you (the authors of the change), the maintainers or the wider community either so I can't win.

PR #469, opened 2021-11-02
Bump sdbus-c++ to version v1.0.0

sangelovic:
There is a new major version of sdbus-c++. This version, among other things, makes some local patches here unnecessary.


kraj (changes requested):
seeing this 

```
ERROR: Nothing RPROVIDES 'gmock' (but /home/jenkins/oe/world/yoe/sources/meta-openembedded/meta-oe/recipes-core/sdbus-c++/sdbus-c++_1.0.0.bb RDEPENDS on or otherwise requires it)
NOTE: Runtime target 'gmock' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['gmock']
NOTE: Runtime target 'sdbus-c++-dev' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['sdbus-c++-dev', 'gmock']
NOTE: Runtime target 'sdbus-c++' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['sdbus-c++', 'gmock']
```
perhaps `gmock` should be added as build dep and not rdep

kraj:
I have fixed it locally and accepted it b26b66e5da

PR #476, opened 2021-11-09
python3-fasteners: replace colon by underscore

shagu:
The dunfell branch can't yet handle the colon notation, which results in an
error while bitbake is parsing recipes files:

  unparsed line: 'RDEPENDS:${PN} += "    ${PYTHON_PN}-logging     ${PYTHON_PN}-fcntl "'

Replacing it by an underscore solves that error.

### DIFF
--- a/meta-python/recipes-devtools/python/python3-fasteners_0.16.3.bb
+++ b/meta-python/recipes-devtools/python/python3-fasteners_0.16.3.bb
@@ -8,7 +8,7 @@ SRC_URI[sha256sum] = "b1ab4e5adfbc28681ce44b3024421c4f567e705cc3963c732bf1cba334
 
 inherit pypi setuptools3
 
-RDEPENDS:${PN} += "\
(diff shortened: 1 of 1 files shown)


shagu:
Nevermind. 

My dunfell branch was not up to date.

Tge support for `:` was brought into `dunfell` with those two commits:
https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?h=dunfell&id=9c2186108e2e8fcf5837c0ffe9dbdeeaea4b7398
https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?h=dunfell&id=1f7b04355c032cd0de4cbe9352745920964c8057

As this recipe is the only one in dunfell that makes use of it, I thought it would be wrong:
```
$ grep -r "RDEPENDS:" *
meta-python/recipes-devtools/python/python3-fasteners_0.16.3.bb:RDEPENDS:${PN} += "\
```

My bad.

PR #481, opened 2021-11-12
libgusb: Use the correct args to disable vala support

hughsie:


hughsie:
@rossburton I think this might help

shr-project:
see https://lists.openembedded.org/g/openembedded-devel/message/93909
which is already in master-next:
https://git.openembedded.org/meta-openembedded/commit/?h=master-next&id=8e816f6fd17ae4142e67b2224ef88dae97435a1c

hughsie:
@shr-project ahh! sorry! Is there anywhere that describes the master-next relationship to master? I was a bit confused what the development was being done. Thanks!

hughsie:
Ross filled me in; I'll submit patches via the mailing list next time. Thanks.

PR #482, opened 2021-11-13
Use old override syntax

otski:
- can-utils and dstat recipes were using the colon override syntax
  introduced in honister
- revert back to underline override notation


shr-project:
In the meantime update your bitbake, the new syntax is backwards compatible all the way to bitbake 1.46 used in dunfell.

akuster:
Can you signoff your change?

otski:
> Can you signoff your change?

Certainly

PR #485, opened 2021-11-16
php: use qemu wrapper to build minilua

meinemitternacht:
minilua, by default, is compiled by the host machine as a
dependency for compiling the JIT/opcache extension. To
successfully cross-compile, this needs to be run under QEMU to
match the target architecture.

According to the [RFC](https://wiki.php.net/rfc/jit), only
x86 and x86_64 targets are supported.

Signed-off-by: Ashley Cox <ashleyc@cybernetics.com>
Signed-off-by: Claude Bing <cbing@cybernetics.com>


kraj:
applied with 265bab83c9

PR #486, opened 2021-11-17
lvm2: do not install systemd units/initscripts when building native SDK tools

bluca:
Nothing picks them up, so the build fails. Required to be able to build
native tools like systemd-analyze.

Signed-off-by: Luca Boccassi <luca.boccassi@microsoft.com>


kraj:
applied with defbcad66f

PR #497, opened 2021-12-02
Use old override syntax for dunfell

nicheath:
https://github.com/openembedded/meta-openembedded/blob/dunfell/meta-python/recipes-devtools/python/python3-fasteners_0.16.3.bb

My dunfell bitbake cannot parse this file after updating to the latest changeset because it is using the new override format.


akuster:
hmm, odd my builders didn't complain.  good catch

nicheath:
I am running bitbake 1.46.0 and the recipe would not parse.

shr-project:
@nicheath is there any reason not to update to newer commit?

nicheath:
I don't think so, I will have to talk with my team, but it is a little surprising to have differing behavior with the same version.

nicheath:
we'll be updating.

PR #501, opened 2021-12-06
libcamera: Update to latest commit

maurosanjo:
Drop patch for musl/aarch64 and bump source commit to latest version that
has the patch applied.

Signed-off-by: Mauro Anjo <maurosanjo@gmail.com>


kraj:
applied with b450b0bb82

PR #513, opened 2022-01-10
Meta python merge

threexc:


kraj:
I have merged all  but https://github.com/openembedded/meta-openembedded/pull/513/commits/c9d34865275bfaf2c0c9be52dea1a8202d75972e
since there is a v2 of this patch, which is now staged in master-next

PR #519, opened 2022-01-18
Add recipe for duktape

bluca:
Embedded javascript engine, work is almost finished to have polkit
use it instead of mozjs, so add a recipe to be ready to switch:
https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/97


kraj:
added with cbbb8ccc75

PR #520, opened 2022-01-18
Meta python merge

threexc:
Missed these two patches in my automatic pipeline. Some builds are breaking without them.


kraj:
merged

PR #535, opened 2022-02-22
abseil-cpp: Fix build with gcc11

apbr:
To use abseil in other recipes it has to be targeted for C++17.


kraj:
this fails grpc see https://errors.yoctoproject.org/Errors/Details/636128/

kraj (changes requested):
fails other dependent packages

apbr:
Strange. Then sorry for bothering.

PR #567, opened 2022-04-21
sdbus-c++-libsystemd: bugfix dev package is not installed

karthago1:
since sdbus-c++-libsystemd outputs only static libraries, it will
not be installed inside the SDK.

add RDEPENDS:${PN}-dev = "" to remove the dependency to the empty package
when installing the sdbus-c++-libsystemd-dev into the SDK

without this change the SDK is missing sdbus-c++-dev and sdbus-c++-libsystemd-dev

Signed-off-by: Ben Fekih, Hichem <hichem.benfekih@ifm.com>

### DIFF
--- a/meta-oe/recipes-core/sdbus-c++/sdbus-c++-libsystemd_250.3.bb
+++ b/meta-oe/recipes-core/sdbus-c++/sdbus-c++-libsystemd_250.3.bb
@@ -59,6 +59,8 @@ EXTRA_OEMESON += "-Dstatic-libsystemd=pic"
 
 S = "${WORKDIR}/git"
 
+RDEPENDS:${PN}-dev = ""
+
 do_compile() {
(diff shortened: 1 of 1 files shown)


kraj:
can you try using `RDEPENDS:${PN}-dev = ""` instead ?

karthago1:
@kraj yes it works also, is there any reason to use `RDEPENDS:${PN}-dev = ""` over `ALLOW_EMPTY:${PN} = "1" ` ?

shr-project:
As explained many times in various places ALLOW_EMPTY just adds completely useless empty package into the image. And creates confusion when someone "installs" the package on target just to discover that the package manager downloaded and installed useless thing which was created just to keep default runtime dependencies in recipes happy.

karthago1:
thank you for the explanation.

kraj:
applied with ec6d073518

PR #577, opened 2022-05-30
Add recipe for Perfetto

quadpixels:
Hello,

This is the rough shape of a BitBake recipe for the [Perfetto](https://perfetto.dev/) performance profiler/visulizer. Perfetto already works for a few OSes on a few architectures such as X86, ARM32 and ARM64, and we recently tested it with OpenBMC (which is mostly ARM).

I'm not super fluent with BitBake, and had done a few ugly steps to make the compilation process pass (I have only tested for OpenBMC-based systems:)

I would like to have some comments on:
* I need to execute the `tools/install-build-deps` script during a Yocto stage that allows network connection
  * Need to make the `curl` tool available to Bitbake
  * Need to make the CA bundle available to Bitbake so curl can access the HTTPS URL
* Because `tools/install-build-deps` is run from the Git working directory, I had to override the unpack step too
* Perfetto builds by generating Ninja scripts with `gn`, and I manually edited the Ninja scripts with a few shell commands, because `gn` only generates certain combinations (linux/macos/windows/solaris, etc + x86/arm/arm64, etc)
* I overrode the `package_qa` task with a no-op to silence the GNU_HASH error

Any comments would be greatly appreciated.
Thanks!

### DIFF
--- a/meta-oe/recipes-devtools/perfetto/perfetto.bb
+++ b/meta-oe/recipes-devtools/perfetto/perfetto.bb
@@ -1,0 +1,151 @@
+LICENSE = "Apache-2.0 & BSD-3-Clause & MIT & Zlib"
+
+LIC_FILES_CHKSUM = "file://LICENSE;md5=f87516e0b698007e9e75a1fe1012b390 \
+                    file://buildtools/libcxx/LICENSE.TXT;md5=55d89dd7eec8d3b4204b680e27da3953 \
+                    file://buildtools/libcxx/utils/google-benchmark/LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57 \
+                    file://buildtools/libcxxabi/LICENSE.TXT;md5=7b9334635b542c56868400a46b272b1e \
+                    file://buildtools/libunwind/LICENSE.TXT;md5=f66970035d12f196030658b11725e1a1 \
+                    file://buildtools/protobuf/LICENSE;md5=37b5762e07f0af8c74ce80a8bda4266b \
+                    file://buildtools/zlib/LICENSE;md5=f09575dbfb09420642318b413159496f \
+                    file://debian/copyright;md5=4e08364c82141f181de69d0a2b89d612 \
+                    file://python/LICENSE;md5=c602a632c34ade9c78a976734077bce7"
+
+SRC_URI = "git://github.com/google/perfetto.git;branch=master;protocol=https;name=perfetto \
+           git://github.com/protocolbuffers/protobuf.git;branch=3.9.x;protocol=https;destsuffix=git/buildtools/protobuf;name=protobuf \
+           git://chromium.googlesource.com/external/github.com/llvm/llvm-project/libcxx.git;protocol=https;destsuffix=git/buildtools/libcxx;branch=main;name=libcxx \
+           git://chromium.googlesource.com/external/github.com/llvm/llvm-project/libcxxabi.git;protocol=https;destsuffix=git/buildtools/libcxxabi;branch=main;name=libcxxabi \
+           git://chromium.googlesource.com/external/github.com/llvm/llvm-project/libunwind.git;protocol=https;destsuffix=git/buildtools/libunwind;branch=main;name=libunwind \
+           git://android.googlesource.com/platform/external/zlib.git;branch=master;protocol=https;destsuffix=git/buildtools/zlib;name=zlib \
+           https://storage.googleapis.com/perfetto/gn-linux64-1968-0725d782;subdir=git/buildtools/;name=gn \
+           file://0001-Remove-check_build_deps-build-steps.patch "
+
+SRCREV_perfetto = "5bd3f582c075d0d026c5fe0b5e291d34dee0d976"
+SRCREV_protobuf = "6a59a2ad1f61d9696092f79b6d74368b4d7970a3"
+SRCREV_libcxx = "d9040c75cfea5928c804ab7c235fed06a63f743a"
+SRCREV_libcxxabi = "196ba1aaa8ac285d94f4ea8d9836390a45360533"
+SRCREV_libunwind = "d999d54f4bca789543a2eb6c995af2d9b5a1f3ed"
+SRCREV_zlib = "5c85a2da4c13eda07f69d81a1579a5afddd35f59"
+SRC_URI[gn.sha256sum] = "f706aaa0676e3e22f5fc9ca482295d7caee8535d1869f99efa2358177b64f5cd"
+
+DEPENDS += " ninja-native"
+
+COMPATIBLE_HOST = "(i.86|x86_64|aarch64|arm).*-linux*"
+
+FILES:${PN}:append = " \
+  ${bindir}/tracebox \
+  "
+
+S = "${WORKDIR}/git"
+B = "${WORKDIR}/build"
+
+# Run the GN (Generate Ninja) script, and replace the compiler flags where applicable
+do_configure () {
+    # Configuration needs to be done from the source directory
+    cd ${S}
+    # Rename a few build tools if they have not been renamed
+    cd buildtools
+    x="gn-linux64-1968-0725d782"
+    [ -f $x ] && mkdir linux64 && mv $x linux64/gn
+    chmod +x linux64/gn
+    cd ..
+
+    CC_BIN=`echo $CC | awk '{print $1}'`
+    CXX_BIN=`echo $CXX | awk '{print $1}'`
+    STRIP_BIN=`echo $STRIP | awk '{print $1}'`
+
+    ARGS="is_debug=false "  # Tell gn to use release mode
+
+    if [ -z `echo ${TOOLCHAIN} | grep clang` ]; then
+        ARGS=$ARGS" is_clang=false"
+    else
+        ARGS=$ARGS" is_clang=true"
+    fi
+    
+    # Architecture parameter accepted by Perfetto
+    arch=${TARGET_ARCH}
+    if [ $arch = "i686" ]; then
+        arch="x86"
+    elif [ $arch = "x86_64" ]; then
+        arch="x64"
+    elif [ $arch = "aarch64" ]; then
+        arch="arm64"
+    fi
+    
+    # For ARM32 with hardware floating point using clang and musl, we need to
+    # specify -mfloat-abi=hard to make the ABI settings of the linker and the
+    # compiler match. The linker would use hardware float ABI. The compiler does
+    # not. As a result we need to force the compiler to do so by adding
+    # -mfloat-abi=hard to compilation flags.
+    FLOAT_ABI=""
+    if [[ "${@bb.utils.contains('TUNE_FEATURES', 'callconvention-hard', 'true', 'false', d)}" == "true" ]]; then
+      FLOAT_ABI="-mfloat-abi=hard"
+    fi
+
+    ARGS=$ARGS" target_os=\"linux\""
+    ARGS=$ARGS" target_cpu=\"$arch\""
+    ARGS=$ARGS" target_cc=\"$CC_BIN ${FLOAT_ABI}\""
+    ARGS=$ARGS" target_cxx=\"$CXX_BIN -std=c++11 ${FLOAT_ABI}\""
+    ARGS=$ARGS" target_strip=\"$STRIP_BIN\"" #
+    ARGS=$ARGS" target_sysroot=\"${RECIPE_SYSROOT}\""
+    ARGS=$ARGS" target_linker=\"$CC_BIN ${FLOAT_ABI} ${LDFLAGS}\""
+    ARGS=$ARGS" target_ar=\"$AR\""
+    ARGS="'$ARGS'"
+    cmd="tools/gn gen --args=$ARGS ${B}"
+
+    echo $cmd
+    # Use eval, not just call $cmd, due to escaping of single quotation marks
+    eval $cmd
+
+    cd ${B}
+    # Eliminate a few incompatible build flags
+    REPLACES="s/-Wl,--icf=all//g"
+    REPLACES=$REPLACES";s/-Werror//g"
+    REPLACES=$REPLACES";s/-mfpu=neon//g"
+    REPLACES=$REPLACES";s/-fcolor-diagnostics//g"
+    REPLACES=$REPLACES";s/=format-security//g"
+    REPLACES=$REPLACES";s/-fdiagnostics-show-template-tree//g"
+    REPLACES=$REPLACES";s/-D_FORTIFY_SOURCE=2//g"
+    REPLACES=$REPLACES";s/-fuse-ld=\S*//g"
+
+    find . -name "*.ninja" | xargs sed $REPLACES -i
+
+    # If using the clang toolchain: use the clang host-side binaries built by Bitbake
+    if [ "${TOOLCHAIN}" = "clang" ]; then
+        BB_CLANGXX="${BUILD_CXX} ${BUILD_LDFLAGS} ${FLOAT_ABI}"
+        BB_CLANG="${BUILD_CC} ${FLOAT_ABI}"
+        BB_LLVM_OBJCOPY="${RECIPE_SYSROOT_NATIVE}/usr/bin/llvm-objcopy"
+        
+        HOST_CLANGXX="${STAGING_DIR_NATIVE}/usr/bin/clang++ -stdlib=libc++ -rtlib=libgcc -unwindlib=libgcc ${FLOAT_ABI}"
+        HOST_CLANG="${STAGING_DIR_NATIVE}/usr/bin/clang ${FLOAT_ABI}"
+        HOST_LLVM_OBJCOPY="${STAGING_DIR_NATIVE}/usr/bin/llvm-objcopy"
+
+        cd gcc_like_host
+        REPLACES="s:\S*clang++ :$HOST_CLANGXX :g"
+        REPLACES=$REPLACES";s:\S*clang :$HOST_CLANG :g"
+        REPLACES=$REPLACES";s:\S*llvm-objcopy :$HOST_LLVM_OBJCOPY :g"
+        find . -name "*.ninja" | xargs sed "$REPLACES" -i
+        cd ..
+    fi
+    # Done processing the Ninja files
+
+    # Workaround for the functions not supported by musl
+    if [ "${TCLIBC}" = "musl" ]; then
+        sed -e 's/strtoll_l(__a, \&__p2, __base, _LIBCPP_GET_C_LOCALE)/strtoll(__a, \&__p2, __base)/g' \
+            -e 's/strtoull_l(__a, \&__p2, __base, _LIBCPP_GET_C_LOCALE)/strtoull(__a, \&__p2, __base)/g' \
+            ${S}/buildtools/libcxx/include/locale -i
+    fi
+}
+
+# Perfetto generates a few different binaries, such as traced and traced_probes and perfetto.
+# The "tracebox" is a busybox that combines the 3 above and provides a single stop for trace capture, so we only build "tracebox" here.
+do_compile () {
+    cd ${B}
+    ninja -C . tracebox
+}
+
+do_install () {
+    BIN_DIR=${D}${bindir}
+
+    install -d -m0755 $BIN_DIR
+    install ${B}/tracebox $BIN_DIR/tracebox
+}
--- a/meta-oe/recipes-devtools/perfetto/files/0001-Remove-check_build_deps-build-steps.patch
+++ b/meta-oe/recipes-devtools/perfetto/files/0001-Remove-check_build_deps-build-steps.patch
@@ -1,0 +1,71 @@
+From 3b7091243ec03054ca8800b51b85a1c09e7e3075 Mon Sep 17 00:00:00 2001
+From: Sui Chen <suichen6@gmail.com>
+Date: Mon, 13 Jun 2022 17:46:49 +0000
+Subject: [PATCH] Remove "check_build_deps" build steps
+
+When building with Bitbake, we want Bitbake to manage our dependencies,
+so we remove the "check_build_deps" steps from Perfetto.
+
+Also setting "is_cross_compiling" to true, so that the host-side tools
+(such as protoc) will always be generated using the native toolchain,
+rather than the Bitbake-generated one.
+
+Signed-off-by: Sui Chen <suichen6@gmail.com>
+---
+ gn/BUILD.gn                  |  1 -
+ gn/standalone/BUILD.gn       | 12 ++++++------
+ gn/standalone/BUILDCONFIG.gn |  5 +++--
+ 3 files changed, 9 insertions(+), 9 deletions(-)
+
+diff --git a/gn/BUILD.gn b/gn/BUILD.gn
+index 8a7ca72a8..e4a2d39f0 100644
+--- a/gn/BUILD.gn
++++ b/gn/BUILD.gn
+@@ -111,7 +111,6 @@ group("default_deps") {
+   deps = [ ":gen_buildflags" ]
+   if (perfetto_build_standalone) {
+     public_deps = [
+-      "//gn/standalone:check_build_deps",
+       "//gn/standalone/libc++:deps",
+       "//gn/standalone/sanitizers:deps",
+     ]
+diff --git a/gn/standalone/BUILD.gn b/gn/standalone/BUILD.gn
+index 582e9b867..9c77ac64c 100644
+--- a/gn/standalone/BUILD.gn
++++ b/gn/standalone/BUILD.gn
+@@ -441,10 +441,10 @@ config("android_liblog") {
+ }
+ 
+ # Checks that tools/install-build-deps has been run since it last changed.
+-perfetto_check_build_deps("check_build_deps") {
+-  args = []
+-}
++#perfetto_check_build_deps("check_build_deps") {
++#  args = []
++#}
+ 
+-perfetto_check_build_deps("check_build_deps_android") {
+-  args = [ "--android" ]
+-}
++#perfetto_check_build_deps("check_build_deps_android") {
++#  args = [ "--android" ]
++#}
+diff --git a/gn/standalone/BUILDCONFIG.gn b/gn/standalone/BUILDCONFIG.gn
+index 6f32686c1..c041989b0 100644
+--- a/gn/standalone/BUILDCONFIG.gn
++++ b/gn/standalone/BUILDCONFIG.gn
+@@ -59,8 +59,9 @@ declare_args() {
+   # the ossfuzz sanitizer overrides this to true. In that config the
+   # host/target cpu and arch are identical, but we want to build only the
+   # targets with the sanitizer/fuzzer flags
+-  is_cross_compiling =
+-      target_cpu != host_cpu || target_os != host_os || target_triplet != ""
++
++  # for Bitbake build
++  is_cross_compiling = true
+ }
+ default_configs = [
+   "//gn/standalone:debug_symbols",
+-- 
+2.37.1
+


meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,145 @@
...
+# To avoid the GNU_HASH warning
+TARGET_CC_ARCH += "${LDFLAGS}"
+
+S = "${WORKDIR}/git"
+
+FILES:${PN}:append = " \
+  ${bindir}/tracebox \
+  "
+
+# Perfetto includes an "install-build-deps" script. It requires network connection.
```
kraj, line 39: this should be perhaps improved. We need to express the SRC_URIs to include all thats needed. The current do_fetch seems a bit clunky

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,145 @@
...
+do_unpack() {
+    tar xfz ../../../../../downloads/perfetto.tar.gz
+    [ -d git ] && rm -r -f git
+    mv perfetto git
+}
+
+# Run the GN (Generate Ninja) script using the Linux-Arm configuration
+# and replace the compiler flags
+do_configure () {
+    SYSROOT=`echo $CC | grep -oi "\-\-sysroot=.*" | sed -s "s/--sysroot=//g"`
```
kraj, line 85: there is a bitbake variable containing this value called RECIPE_SYSROOT

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,145 @@
...
+    BUILD_DIR=${WORKDIR}"/build"
```
kraj, line 86: Use B instead of defining duplicate BUILD_DIR

kraj (changes requested):

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Done.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: It seems `B` evaluates to `/opt/evb-ast2500/tmp/work/arm1176jzs-openbmc-linux-gnueabi/perfetto/1.0-r0/git` on my system, so I would need to use ${B}/../build and that may not be as good as BUILD_DIR.

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,145 @@
...
+    
+    echo "Installing dependencies"
+    tools/install-build-deps
+
+    CC_BIN=`echo $CC | awk '{print $1}'`
+    CXX_BIN=`echo $CXX | awk '{print $1}'`
+    STRIP_BIN=`echo $STRIP | awk '{print $1}'`
+
+    ARGS="is_debug=false "  # Tell gn to use release mode
+    ARGS=$ARGS" is_clang=false"
```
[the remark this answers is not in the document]
quadpixels: Done. Tested by setting `TOOLCHAIN=clang`, the build process completes.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Could you explain does `express the SRC_URIs to include all thats needed` mean? I can also look into other recipes to see how this is handled.

It is now clunky because I need to run `python3 tools/install-build-deps` in the fetch step (which requires network connection, so it has to be in the fetch step), and have Bitbake pack the source tree plus the installed dependencies into an archive.

shr-project: But calling this script in do_fetch isn't the same as bitbake fetcher fetching it from SRC_URI. Because bitbake fetcher will respect MIRROR/PREMIRROR variables and store the archives in DL_DIR, so next time you build this recipe it won't need to reach the network and just unpack from local DL_DIR.

You need to replicate what this script does (copy the URLs and revision etc) and then persuade the build that the sources are already downloaded (sometimes it's easy and the script just checks that the deps are there, sometimes there are extra variables you can set and sometimes you have to patch the build to allow this behavior. I haven't looked at what perfetto does exactly, but I had to deal with similar issues in meta-ros quite often, you can find some examples there.

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,145 @@
...
+        python3 tools/install-build-deps
+        cd ..
+        tar cfz perfetto.tar.gz perfetto
+        touch perfetto.tar.gz.done
+        rm -r -f perfetto
+    fi
+}
+
+do_unpack() {
+    tar xfz ../../../../../downloads/perfetto.tar.gz
```
shr-project, line 77: This won't work for with any modification to DL_DIR or TMPDIR, at least use DL_DIR but both do_fetch and do_unpack should be simplified as mentioned above.

quadpixels: Thanks for the suggestion, I replaced `../../../../../downloads` with DL_DIR.

Is the "simplified" referring to "express the SRC_URIs to include all thats needed"? I'm trying to figure out how to do it.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Some small updates: I have made some progress after realizing `SRC_URI` can contain multiple sources/repositories. I'm now trying to figure out how to properly do the "unpack" step. The default unpack seems to overwrite whatever has already been "unpacked" so I'm looking for a way to achieve the desired unpack behavior (copy all dependencies like googletest, protobuf, libcxxabi, etc. into Perfetto.)

quadpixels: So I asked around on the OpenBMC Discord channel and managed to set SRC_URI and eliminated the custom fetch and unpack steps.

meta-oe/recipes-devtools/perfetto/perfetto.bb
shr-project: You can set B = "${WORKDIR}/build" as benefit bitbake will automatically cd to ${B} for configure/compile/install so you don't have to.

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,151 @@
...
+do_install () {
+    BUILD_DIR=${WORKDIR}/build/
+    BIN_DIR=${D}${bindir}
+
+    install -d -m0755 $BIN_DIR
+    install $BUILD_DIR/tracebox $BIN_DIR/tracebox
+}
+
+do_package_qa() {
+    echo "Skipping package_qa to silence the GNU_HASH error"
```
shr-project, line 150: Use INSANE_SKIP variable if it's too difficult to pass correct LDFLAGS into the build, no need to skip the whole function.

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,151 @@
+LICENSE = "Apache-2.0 & Unknown"
```
shr-project, line 1: Which parts have unknown license?

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,151 @@
...
+SRCREV_perfetto = "5bd3f582c075d0d026c5fe0b5e291d34dee0d976"
+SRCREV_googletest = "609281088cfefc76f9d0ce82e1ff6c30cc3591e5"
+SRCREV_protobuf = "6a59a2ad1f61d9696092f79b6d74368b4d7970a3"
+SRCREV_libcxx = "d9040c75cfea5928c804ab7c235fed06a63f743a"
+SRC_URI[libcxx.sha256sum] = "c49ecee61ac94784a08ec9ecd5248512d0899ce88b26a90b9406142d93458c63"
+SRCREV_libcxxabi = "196ba1aaa8ac285d94f4ea8d9836390a45360533"
+SRCREV_libunwind = "d999d54f4bca789543a2eb6c995af2d9b5a1f3ed"
+SRCREV_fuzzer = "debe7d2d1982e540fbd6bd78604bf001753f9e74"
+SRCREV_benchmark = "e991355c02b93fe17713efe04cbc2e278e00fdbd"
+SRCREV_libbacktrace = "177940370e4a6b2509e92a0aaa9749184e64af43"
```
shr-project, line 49: You don't need both SRCREV and SRC_URI checksum, right as you're not fetching libbacktrace with git fetcher.

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,151 @@
...
+           git://chromium.googlesource.com/chromium/llvm-project/cfe/tools/clang-format.git;protocol=https;destsuffix=git/buildtools/clang-format;name=clang-format \
+           "
+
+# Todo: if toolchain is clang, add clang SRC_URI's, or make Perfetto use Yocto's clang
+
+SRCREV_perfetto = "5bd3f582c075d0d026c5fe0b5e291d34dee0d976"
+SRCREV_googletest = "609281088cfefc76f9d0ce82e1ff6c30cc3591e5"
+SRCREV_protobuf = "6a59a2ad1f61d9696092f79b6d74368b4d7970a3"
+SRCREV_libcxx = "d9040c75cfea5928c804ab7c235fed06a63f743a"
+SRC_URI[libcxx.sha256sum] = "c49ecee61ac94784a08ec9ecd5248512d0899ce88b26a90b9406142d93458c63"
```
shr-project, line 44: This one doesn't seem to be used.

shr-project (changes requested):

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: It turns out passing the correct LDFLAGS is pretty easy, so this custom do_package_qa is removed.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: I checked the licenses and updated the license section accordingly:
* Perfetto: Apache-2.0
* benchmark: Apache-2.0
* googletest: BSD-3-Clause
* jsoncpp: MIT
* jsoncpp/devtools/licenseupdater.py: MIT
* libcxx: Apache-2.0
* libcxx/utils/google-benchmark: Apache-2.0
* libcxxabi: Apache-2.0
* libunwind: Apache-2.0
* linenoise: BSD-2-Clause
* protobuf: BSD-3-Clause
* sqlite: Public Domain
* zlib: ZLib
* Debian Perfetto package license: Apache-2.0
* Python module license: Apache-2.0

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: It seems BitBake still requires a "SRCREV_libbacktrace" even if it's fetched as a zip package, otherwise the following error appears:
```
ERROR: perfetto-1.0-r0 do_fetch: No checksum specified for '/opt/evb-ast2500/downloads/libbacktrace-177940370e4a6b2509e92a0aaa9749184e64af43.zip', please add at least one to the recipe:
SRC_URI[libbacktrace.sha256sum] = "21ac9a4209f7aeef766c482be53a7fa365063c031c7077e2070b491202983b31"
ERROR: perfetto-1.0-r0 do_fetch: Bitbake Fetcher Error: NoChecksumError('Missing SRC_URI checksum', 'https://storage.googleapis.com/perfetto/libbacktrace-177940370e4a6b2509e92a0aaa9749184e64af43.zip;subdir=git/buildtools;name=libbacktrace')
```

shr-project: Are you sure it was after removing "SRCREV_libbacktrace"? The error looks like SRC_URI[libbacktrace.sha256sum] was removed.

quadpixels: I realized the "srcrev" was a left-over typo when I found I could not fetch using the name of their git repository. So yes you are right the error is due to the lack of the SHA256 sum.

I tried all 4 possibilities for the libbacktrace.zip:

1. No sha256sum, no srcrev:
Error "no checksum specified"

2. Only sha256sum, no srcrev:
Fetch OK

3. No sha256sum, only srcrev:
Error "no checksum specified"

4. Both sha256sum and srcrev:
Fetch OK

Also tried all 4 possibilities for the libcxx repository:

1. No sha256sum, no srcrev:


2. Only sha256sum, no srcrev:
Error "please set a valid SRCREV for URI"

3. No sha256sum, only srcrev:
Fetch OK

4. Both sha256sum and srcrev:
Fetch OK

So a SHA256 sum is required for a file URI, while srcrev is required for a repository.

shr-project: Yes, checksums are needed for http(s) fetcher, SRCREVs for git fetcher.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Agreed, B is now set to `"${WORKDIR}/build"` to make the script look more clear.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: This is not relevant now because the manual unpack step is eliminated.

quadpixels:
Modified the recipe a bit so it works with `devtool extract` and `devtool modify -n`. The main points is to use `${S}` so Bitbake enters the source directory for configuration (it must enter the source directory to configure Perfetto due to Perfetto works now).
Also change the binary file from the un-stripped one (which is ~52MB in size) to the stripped one (~1MB in size).

Please take another look. Thanks!

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,116 @@
...
+SRCREV_libcxxabi = "196ba1aaa8ac285d94f4ea8d9836390a45360533"
+SRCREV_libunwind = "d999d54f4bca789543a2eb6c995af2d9b5a1f3ed"
+SRCREV_zlib = "5c85a2da4c13eda07f69d81a1579a5afddd35f59"
+SRC_URI[gn.sha256sum] = "f706aaa0676e3e22f5fc9ca482295d7caee8535d1869f99efa2358177b64f5cd"
+
+DEPENDS += " ninja-native"
+
+B = "${WORKDIR}/build"
+
+FILES:${PN}:append = " \
```
kraj, line 34: just use `+=` instead of `:append`

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,116 @@
...
+SRCREV_protobuf = "6a59a2ad1f61d9696092f79b6d74368b4d7970a3"
+SRCREV_libcxx = "d9040c75cfea5928c804ab7c235fed06a63f743a"
+SRCREV_libcxxabi = "196ba1aaa8ac285d94f4ea8d9836390a45360533"
+SRCREV_libunwind = "d999d54f4bca789543a2eb6c995af2d9b5a1f3ed"
+SRCREV_zlib = "5c85a2da4c13eda07f69d81a1579a5afddd35f59"
+SRC_URI[gn.sha256sum] = "f706aaa0676e3e22f5fc9ca482295d7caee8535d1869f99efa2358177b64f5cd"
+
+DEPENDS += " ninja-native"
+
+B = "${WORKDIR}/build"
```
kraj, line 32: this is redundant

kraj:
also getting patching errors 

```
ERROR: perfetto-1.0-r0 do_patch: Applying patch '0001-Remove-check_build_deps-build-steps.patch' on target directory '/mnt/b/yoe/master/build/tmp/work/cortexa15t2hf-neon-yoe-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0'
CmdError('quilt --quiltrc /mnt/b/yoe/master/build/tmp/work/cortexa15t2hf-neon-yoe-linux-gnueabi/perfetto/1.0-r0/recipe-sysroot-native/etc/quiltrc push', 0, 'stdout: Applying patch 0001-Remove-check_build_deps-build-steps.patch
can\'t find file to patch at input line 17
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|From ec414dc44702c71db3d57d69b27a388bc3426cad Mon Sep 17 00:00:00 2001
|From: Sui Chen <suichen@google.com>
|Date: Mon, 13 Jun 2022 17:46:49 +0000
|Subject: [PATCH] Remove "check_build_deps" build steps
|
|When building with Bitbake, we want Bitbake to manage our dependencies,
|so we remove the "check_build_deps" steps from Perfetto.
|---
| gn/BUILD.gn            |  1 -
| gn/standalone/BUILD.gn | 12 ++++++------
| 2 files changed, 6 insertions(+), 7 deletions(-)
|
|diff --git a/gn/BUILD.gn b/gn/BUILD.gn
|index 8a7ca72a8..e4a2d39f0 100644
|--- a/gn/BUILD.gn
|+++ b/gn/BUILD.gn
--------------------------
No file to patch.  Skipping patch.
1 out of 1 hunk ignored
can\'t find file to patch at input line 29
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/gn/standalone/BUILD.gn b/gn/standalone/BUILD.gn
|index 33fd7aab2..6235044c7 100644
|--- a/gn/standalone/BUILD.gn
|+++ b/gn/standalone/BUILD.gn
--------------------------
No file to patch.  Skipping patch.
1 out of 1 hunk ignored
Patch 0001-Remove-check_build_deps-build-steps.patch does not apply (enforce with -f)


```

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Done.

meta-oe/recipes-devtools/perfetto/perfetto.bb
quadpixels: Sorry I didn't check this section very well, I did a more detailed test for all possible combinations. It seems I need to set both `S` and `B`:

1. Not setting `S` or `B`, without `devtool override -n perfetto /tmp/perfetto`:
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
--> "patch" task will fail, because source codes are still checked out to ${WORKDIR}/git

2. Not setting S or B, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail (because Perfetto doesn't allow the build folder to be the same as the source folder)

3. Only specifying `S = "${WORKDIR}/git"`, without override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
--> "configure" task will fail (same as above)

4. Only specifying S, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail (same as above)

5. Only specifying `B = "${WORKDIR}/build"`, without override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/build"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
--> "patch" task will fail, because source codes are still checked out to ${WORKDIR}/git

6. Only specifying B, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail

7. Specifying both S and B, no override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/build"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
--> works

8. Specifying both S and B, with override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0/"`
`S="/tmp/perfetto"`
--> works

meta-oe/recipes-devtools/perfetto/perfetto.bb
```diff
@@ -0,0 +1,116 @@
...
+
+    ARGS="is_debug=false "  # Tell gn to use release mode
+
+    if [ -z `echo ${TOOLCHAIN} | grep clang` ]; then
+        ARGS=$ARGS" is_clang=false"
+    else
+        ARGS=$ARGS" is_clang=true"
+    fi
+    ARGS=$ARGS" target_os=\"linux\""
+    ARGS=$ARGS" target_cpu=\"arm\""
```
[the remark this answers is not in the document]
quadpixels: Done.

quadpixels:
Sorry for the confusion about the build error, I didn't check Yocto's source and build folders very closely, and ran into some side-effects of stuck environment variables. I did some more detailed, clean tests for all 8 possible combinations. It seems I need to set both:

Not setting `S` or `B`, without devtool override -n perfetto /tmp/perfetto:
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
--> "patch" task will fail (because source codes are still checked out to `${WORKDIR}/git`)

Not setting `S` or `B`, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail (because Perfetto doesn't allow the build folder to be the same as the source folder)

Only setting `S = "${WORKDIR}/git"`, without override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
--> "configure" task will fail (same as above)

Only specifying `S`, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail (same as above)

Only specifying `B = "${WORKDIR}/build"`, without override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/build"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0"`
--> "patch" task will fail (same as above)

Only specifying B, with override
`B="/tmp/perfetto"`
`S="/tmp/perfetto"`
--> "configure" task will fail (same as above)

Specifying both S and B, no override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/build"`
`S="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/git"`
--> works

Specifying both S and B, with override
`B="/opt/build/tmp/work/armv7a-openbmc-linux-gnueabi/perfetto/1.0-r0/perfetto-1.0/"`
`S="/tmp/perfetto"`
--> works

So with both `S` and `B` variables set in the recipe, the recipe works both with and without a `devtool modify -n` override.

quadpixels:
I updated the script and tested with `qemux86` and `qemux86-64` targets. Both can generate the build files fine. Other architectures, such as mips, do not appear to be supported by Perfetto yet.
I need to force the `is_cross_compiling` flag to true, otherwise the build system would attempt to execute Bitbake-generated intermediate tools like `protoc`. Those generated files generated by Yocto would not work on a regular host due to the version of the C run time. So setting `is_cross_compiling` to true such that the host-side tools are generated using the native toolchain rather than the Bitbake-generated one.

quadpixels (commented):
The referenced files have been updated per request

quadpixels:
Friendly ping, thanks!

kraj:
> Friendly ping, thanks!

 I am busy traveling next week. I will try to get to it this week ( provided my machine recovers )

quadpixels:
Friendly ping for review --- thanks!

kraj:
> Friendly ping for review --- thanks!

thanks, it fell off radar, I will give it another chance.

kraj (changes requested):
fails on musl/arm64
https://errors.yoctoproject.org/Errors/Details/667305/

quadpixels:
> fails on musl/arm64 https://errors.yoctoproject.org/Errors/Details/667305/

Thanks for the quick response, the build error was due to the `current_cpu` variable in the getting a value of "aarch64" while the GN script only expects "arm64". The variable is populated through `arch` in the BB recipe. Added an if statement to replace this value and the build can complete on the qemuarm64 platform.

Is there a way to manually trigger the AutoBuilder for testing?

quadpixels:
Also added Signed-off-By to the patch file. Tested building for qemuarm64 locally.
Are there any other issues that need to be resolved?
Thanks!

kraj:
> Added Signed-off-By to the patch file. Tested building for qemuarm64 locally. Are there any other issues that need to be resolved? Thanks!

I will give it a go on Autobuilder again.

kraj:
> > Added Signed-off-By to the patch file. Tested building for qemuarm64 locally. Are there any other issues that need to be resolved? Thanks!
> 
> I will give it a go on Autobuilder again.

fails with clang compiler this time

https://errors.yoctoproject.org/Errors/Details/668126/

We should be using OE built clang. It seems to be poking at build host

quadpixels:
I was able to reproduce the build error by downloading and configuring a Yoe Distribution.

The clang++ generated by the Yoe seems to only look into system directories:

clang++ (version 15) generated by Yoe:
> `$ git clone --recurse-submodules -j8 -b master https://github.com/YoeDistro/yoe-distro.git yoe`
> `$ yoe_setup`
> `$ bitbake perfetto -c devshell`
> `$ /home/nitroglycerine/yoe/build/tmp/work/cortexa57-yoe-linux/perfetto/1.0-r0/build# /home/nitroglycerine/yoe/build/tmp/work/cortexa57-yoe-linux/perfetto/1.0-r0/recipe-sysroot-native/usr/bin/clang++  /home/nitroglycerine/yoe/a.cpp  -v`
> `...`
> ` "/home/nitroglycerine/yoe/build/tmp/hosttools/ld" -pie --hash-style=both --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o a.out /usr/lib/x86_64-linux-gnu/Scrt1.o /usr/lib/x86_64-linux-gnu/crti.o /usr/lib/gcc/x86_64-linux-gnu/10/crtbeginS.o -L/usr/lib/gcc/x86_64-linux-gnu/10 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/lib -L/usr/lib /tmp/a-4acb41.o -lc++ -lm /usr/lib/clang/15.0.0/lib/linux/libclang_rt.builtins-x86_64.a --as-needed -lunwind --no-as-needed -lc /usr/lib/clang/15.0.0/lib/linux/libclang_rt.builtins-x86_64.a --as-needed -lunwind --no-as-needed /usr/lib/gcc/x86_64-linux-gnu/10/crtendS.o /usr/lib/x86_64-linux-gnu/crtn.o`

clang++ (version 14) generated by OpenBMC distribution:
> `$ bitbake perfetto -c devshell`
> `$ arm-openbmc-linux-gnueabi-clang++ /tmp/a.cpp -v`
> `...`
> ` "/opt/evb-ast2500/tmp/work/arm1176jzs-openbmc-linux-gnueabi/perfetto/1.0-r0/recipe-sysroot-native/usr/bin/arm-openbmc-linux-gnueabi/arm-openbmc-linux-gnueabi-ld" -pie -EL -z relro -X --hash-style=gnu --eh-frame-hdr -m armelf_linux_eabi -dynamic-linker /lib/ld-linux.so.3 -o a.out Scrt1.o crti.o crtbeginS.o -L/opt/evb-ast2500/tmp/work/arm1176jzs-openbmc-linux-gnueabi/perfetto/1.0-r0/recipe-sysroot-native/usr/bin/../lib/arm-openbmc-linux-gnueabi -L//lib/../lib -L//usr/lib/../lib -L/opt/evb-ast2500/tmp/work/arm1176jzs-openbmc-linux-gnueabi/perfetto/1.0-r0/recipe-sysroot-native/usr/bin/../lib -L//lib -L//usr/lib /tmp/a-18b282.o -lstdc++ -lm -lgcc_s -lgcc -lc -lgcc_s -lgcc crtendS.o crtn.o`

For the Yoe distribution, I followed the Yoe tutorial, the only modification I made is to copy `meta-openembedded/meta-oe/recipes-devtools/perfetto` to `~/yoe/sources/meta-openembedded/meta-oe/recipes-devtools`. Is this a close resemblance of the build environment used by the Auto Builder?

quadpixels:
The above error is the error with the native clang generated by Yocto. It happens with the clang-15 bundled with the Yoe distribution, and does not appear in the clang-14 bundled with OpenBMC.
An yet simpler way to reproduce the error:
```
$ cat a.cpp
#include <stdio.h>
int main() { printf("Hey!\n"); }
```
clang-15 from the Yoe distribution:
```
$ ~/yoe/build/tmp/work/x86_64-linux/clang-native/15.0.0-r0/build/bin/clang++ -nostdlib++ /tmp/a.cpp 
/usr/bin/ld: cannot find /usr/lib/clang/15.0.0/lib/linux/libclang_rt.builtins-x86_64.a: No such file or directory
/usr/bin/ld: cannot find -lunwind
```
clang-14 from OpenBMC:
```
$ /opt/evb-ast2500/tmp/work/x86_64-linux/clang-native/14.0.0-r0/build/bin/clang++ /tmp/a.cpp 
(successfully completes)
```
Trying to figure out what happened in the native clang build.

quadpixels:
Forcing clang's compiler runtime to "libgcc" seems to fix the problem with the Yoe build.

```
BB_CLANGXX="${RECIPE_SYSROOT_NATIVE}/usr/bin/clang++ --rtlib=libgcc --unwindlib=libgcc"
```

Please take another look if this is okay or if there is a better solution. Thanks!

quadpixels:
According to the reply in [this TMPDIR post](https://community.tmpdir.org/t/libclang-rt-builtins-x86-64-a-missing-for-native-clang-15/640/2), updated the recipe to inherit the [`clang-native` class](https://community.tmpdir.org/t/libclang-rt-builtins-x86-64-a-missing-for-native-clang-15/640/2) such that it picks up the [`BUILD_LDFLAGS` specified in the class](https://github.com/kraj/meta-clang/blob/master/classes/clang-native.bbclass#L16) that sets clang's runtime to gcc, and makes the build pass.

kraj:
It failed on mips see
https://errors.yoctoproject.org/Errors/Details/672086/

also on musl see
https://errors.yoctoproject.org/Errors/Details/672091/

IMO its best to limit the compatibility to the architectures it supports, its fine to ignore mips.

```
COMPATIBLE_HOST = "(i.86|x86_64|powerpc|powerpc64|aarch64|arm).*-linux*"`
```

or further cut this list to down to where it is tested to work. 

I think the musl failure seems should be fixed.

quadpixels:
Added `COMPATIBLE_HOST` so that only supported platforms are included.
For the musl failure, the cause seems to be:
1) `strtoull_l` and `strtoll_l` are not supported in musl.
2) The `BUILD_CXX` compiler (aarch64-yoe-linux-musl-clang++) is used for certain host-side libraries. 

Fixes:
1) Substitute the two with the ones without the `_l` suffix, namely, `strtoull` and `strtoll`. This is done by running `sed` in the configure step.
2) Use clang++ located in `STAGING_DIR_NATIVE`.

kraj:
I am seeing following errors when not using meta-clang

```
Parsing recipes...ERROR: ParseError at /home/pokybuild/yocto-worker/meta-oe/build/meta-openembedded/meta-oe/recipes-devtools/perfetto/perfetto.bb:31: Could not inherit file classes/clang-native.bbclass

```

I think you should be using dynamic-layers technique to enable it only when when meta-clang is enabled.

See https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/conf/layer.conf#L28

add `clang-layer` there

and then define the recipe under `meta-oe/dynamic-layers/clang-layer/recipes-devtools`

quadpixels:
> I following errors when not using meta-clang
> 
> ```
> Parsing recipes...ERROR: ParseError at /home/pokybuild/yocto-worker/meta-oe/build/meta-openembedded/meta-oe/recipes-devtools/perfetto/perfetto.bb:31: Could not inherit file classes/clang-native.bbclass
> ```

Yes, I noted that error as well, when I was building for an OpenBMC platform without meta-clang enabled.
It seems removing `inherits clang-native` from this recipe fixes this error.

> and then define the recipe under `meta-oe/dynamic-layers/clang-layer/recipes-devtools`

Perfetto supports both building with clang and without clang, and the original purpose of `inherit clang-native` was to make sure the native clang++ compiler is available, so [lines 112~121](https://github.com/openembedded/meta-openembedded/pull/577/files#diff-077f5c135ea5b6a7097438189cbbe8c4ce990d833ead8ad24888ccac33981639R112-R121) of the patch can work normally.
It seems to me when the meta-clang layer is enabled, it is implied that clang-native is present on the system and the native clang++ compiler can be found in `${STAGING_DIR_NATIVE}/usr/bin/clang++`, and the above lines would work correctly, thus there it seems we don't need to inherit clang-native.

kraj:
slowly getting better, there are warnings seen

https://autobuilder.yoctoproject.org/typhoon/#/builders/88/builds/2106

I think you need to specify `branch=` also in SRS_URI

quadpixels:
Added `branch=` to the three repositories: "master" for perfetto, "main" for libcxx, libcxxabi, and libunwind.

There is only one warning left: `WARNING: perfetto-1.0-r0 do_package_qa: QA Issue: File /usr/bin/.debug/tracebox in package perfetto-dbg contains reference to TMPDIR [buildpaths]`.

Trying to find a fix ...

quadpixels:
Was able to quench the "reference to TMPDIR" warning with `INHIBIT_PACKAGE_DEBUG_SPLIT = "1"`, not sure if this is the best approach though.

kraj:
> Was able to quench the "reference to TMPDIR" warning with `INHIBIT_PACKAGE_DEBUG_SPLIT = "1"`, not sure if this is the best approach though.

I am fine to live with a warning and not silence it like this. Because the root cause is that some path is leaking into compilation and that needs to be fixed

quadpixels:
Thanks for the suggestions, I removed the `INHIBIT_PACKAGE_DEBUG_SPLIT` line.

I opened the debug symbol file with "Cutter", it showed 3 strings that contain the TMPDIR. Those strings are in the `.debug_str` section of the binary. The "Cutter" decompiler does seem to work very well with that debug symbol file though, perhaps because it wasn't a proper executable.

I also tried setting `-fcoverage-prefix-map`, `-fdebug-prefix-map`, `-ffile-prefix-map` and `-fmacro-prefix-map` to empty values, this did not fix the QA warning.

kraj:
Search the absolute path of your workspace in the output of strings command that might give some clue.

kraj:
@quadpixels thanks for sticking with me and addressing all review feedback, it has been accepted into master with 8073ec2275

PR #581, opened 2022-06-16
clinfo: use virtual opencl loader provider

jybros:
Hello,

Use **virtual/opencl-icd** as OpenCL provider instead of static Khronos **opencl-icd-loader** (similar to **PACKAGECONFIG[opencl]** found in OpenCV recipe).

One might decide to link straight with a specific OpenCL implementation (GPU-vendor specific, PoCL, ...). In that case one only needs to override **PREFERRED_PROVIDER_virtual/opencl-icd** once and for all.


kraj:
applied to master with 1667aef536

PR #584, opened 2022-06-20
fix(syslog-ng): warning about conf version

### LINKED ISSUE #583: syslog-ng: warning on startup
On kikstone branch, I got this warning when syslog-ng service starts on boot:
```
WARNING: Configuration file format is too old, syslog-ng is running in compatibility mode. Please update it to use the syslog-ng 3.36 format at your time of convenience. To upgrade the configuration, please review the warnings about incompatible changes printed by syslog-ng, and once completed change the @version header at the top of the configuration file; config-version='3.31'
```
I understand the version of the recipe is 3.36.x but the default config file (`syslog-ng.conf.sysvinit` in my case) indicates `@version: 3.31`, hence the warning.
Maybe we should consider incrementing the version in the file to `3.36`.

Nautigsam:
The service warned on startup about running in compatibility mode since the configuration version was "3.31" instead of "3.36".
Fixes #583.

### DIFF
--- a/meta-oe/recipes-support/syslog-ng/files/syslog-ng.conf.systemd
+++ b/meta-oe/recipes-support/syslog-ng/files/syslog-ng.conf.systemd
@@ -1,4 +1,4 @@
-@version: 3.31
+@version: 3.36
 #
 # Syslog-ng configuration file, compatible with default Debian syslogd
 # installation. Originally written by anonymous (I can't find his name)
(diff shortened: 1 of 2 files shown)


Nautigsam:
@akuster Not sure if a PR is acceptable, tell me if you prefer a real patch file.

kraj:
applied to master with e80ce510e1

PR #588, opened 2022-07-19
ttf-fonts: fix URIs, upgrade 1.004 -> 2.004

kazuki0824:


kraj:
applied with 6d05828e4a

PR #591, opened 2022-08-09
lapack: add packageconfig for lapacke

azaharia-wr:
Signed-off-by: Adrian Zaharia <Adrian.Zaharia@windriver.com>


kraj:
applied with d799db35da

PR #594, opened 2022-08-20
Revert "Suppress eol in functionfs setup scripts (#147)"

tewarid:
This reverts commit db5f48734404a52ee5323659082f1d6baa225ca7.

It does not seem to be necessary any more and is impeding startup of adbd on a raspberry pi system built with clang, musl, and toybox.

Signed-off-by: Devendra Tewari <devendra.tewari@gmail.com>

### DIFF
--- a/meta-oe/recipes-devtools/android-tools/android-tools-conf-configfs/android-gadget-cleanup
+++ b/meta-oe/recipes-devtools/android-tools/android-tools-conf-configfs/android-gadget-cleanup
@@ -6,7 +6,7 @@ cd /sys/kernel/config/usb_gadget
 
 cd adb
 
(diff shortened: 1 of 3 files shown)


kraj (commented):
does this work on non toybox system e.g. with busybox ? can you check

tewarid:
doesn't work with busybox :(

closing this, will resubmit later

tewarid:
please don't integrate upstream, the problem isn't the reverted commit. #595 has the appropriate fix for the issue I am seeing.

kraj:
> please don't integrate upstream, the problem isn't the reverted commit. #595 has the appropriate fix for the issue I am seeing.

It has been dropped already thanks

PR #595, opened 2022-08-22
android-tools: sleep more in android-gadget-start

tewarid:
Increase sleep to 10 seconds otherwise interface is not ready on rpi0w.

Verified on rpi0w system built with clang, musl, and toybox and/or busybox.

Signed-off-by: Devendra Tewari <devendra.tewari@gmail.com>


kraj:
applied with 8403930ab4

PR #597, opened 2022-08-24
hiba: Add initial recipe for hiba

wltu:
HIBA is a system built on top of regular OpenSSH certificate-based
authentication that allows to manage flexible authorization of
principals on pools of target hosts without the need to push customized
authorized_users files periodically.

This recipe allow us to manage SSH to production machines easily and not
rely on authorized keys.

The install steps are listed in
https://github.com/google/hiba/blob/main/INSTALL.md

It depends on the `openssh-portable`'s file and object files, so I am
building it in the code source to match the install instruction.

Tested:
Ran the local-setup.sh and works fine on a BMC.
```
$ ssh -F /tmp/tmp.w8itTW/ssh_config -p 2201 root@localhost
 #####################################################
 #                 setup-local.sh                    #
 # Host Identity Based Authorization SSHD example.   #
 #####################################################
prodHost: limited access
Connection to localhost closed.
$ ssh -F /tmp/tmp.w8itTW/ssh_config -p 2202 root@localhost
 #####################################################
 #                 setup-local.sh                    #
 # Host Identity Based Authorization SSHD example.   #
 #####################################################
Last login: Thu Jan  1 00:47:28 1970 from ::1
```


meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,40 @@
+
+SUMMARY = "Google Host Identity Based Authorization"
+DESCRIPTION = "HIBA is a system built on top of regular OpenSSH \
+  certificate-based authentication that allows to manage flexible \
+  authorization of principals on pools of target hosts without the need to \
+  push customized authorized_users files periodically."
+
+S = "${WORKDIR}/git"
+
+LICENSE = "BSD"
```
kraj, line 10: Be specific about which BSD license e.g `BSD-3-Clause`

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,40 @@
...
+  if len(openssh_paths) != 1:
+    raise Exception("openssh_paths does not have only one path: {}".format(openssh_paths))
+  
+  return "--prefix=/usr/ --with-opensshdir={}".format(openssh_paths[0])
+
+EXTRA_OECONF = "${@get_oeconf(d)}"
+
+# Make sure that MACHINE_FEATURES includes hiba before continuing
+do_configure:prepend(){
+  if [[ "${MACHINE_FEATURES}" != *"hiba"* ]]; then
```
kraj, line 36: No bashism please, we do support other shells e.g. dash for parsing generated task scripts

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,40 @@
...
+def get_oeconf(d):
+  import glob
+  workdir = d.getVar("WORKDIR")
+  openssh_paths = glob.glob("{}/../../openssh/*/openssh-*".format(workdir))
+  if len(openssh_paths) != 1:
+    raise Exception("openssh_paths does not have only one path: {}".format(openssh_paths))
+  
+  return "--prefix=/usr/ --with-opensshdir={}".format(openssh_paths[0])
+
+EXTRA_OECONF = "${@get_oeconf(d)}"
```
kraj, line 32: I feel this should be avoided. Can you explain what you are trying to do with thi s function ?

kraj (changes requested):

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,40 @@
...
+
+inherit autotools
+
+DEPENDS = "openssl openssh zlib"
+RDEPENDS:${PN} = "bash"
+
+def get_oeconf(d):
+  import glob
+  workdir = d.getVar("WORKDIR")
+  openssh_paths = glob.glob("{}/../../openssh/*/openssh-*".format(workdir))
```
wak-google, line 26: I don't believe you can rely on directories above WORKDIR within the recipe.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: In https://github.com/google/hiba/blob/main/INSTALL.md, the hiba repo depends on the headers/object files from openssh to build. So we need to the build folder from openssh to buidl hiba and that path cannot be hardcoded.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: oh dang, I am not too sure what else to use then.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Updated. Is this better? Or did I do it wrong again?

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Done.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
BenjaminFair: I'm guessing you'll need to modify the openssh recipe instead to make the files that you need available

wltu: where should I install the files? William mentioned that we shouldn't use the `STAGING_DIR_NATIVE` which I thought is the sysroot. (maybe not)

kraj: if you are using gut of a package ( e.g. openssh ) without respecting what the component expect their users to use via headers and libs then I would urge you to take a look at this component and try to either use what openssh exposes or make it expose what you need. Yocto builds in recipe specific sysroots, they dont see anything outside that sysroot during build. As such we have shared state and this will break horribly if their are dependencies outside its own sysroot.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: It seems like `openssh-dev` and `openssh-staticdev` are both empty. It seems like the header and object files are not exposed anywhere. So it sounds like this is not doing without the "hack" that exists right now.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: um, so it sounds like we could not build `hiba` in bitbake since it depends on stuff that is not exposed.

I guess this means that the only way to do it cleanly is to have the `hiba` as a submodule of `openssh-portable` to enforce that all the required stuff are available... but then that will be dependent on the openssh release to get the new changes... which is not ideal.

@blundere what do you think? Seems like the way it's done is not proper... and I have ran out of ideas.

kraj: as a submod of openssh seems would work and sounds right too. You can use a `PACKAGECONFIG` knob then to control turning it on/off

blunderer: As I understand it, we need to be modifying the OpenSSH recipe to expose the headers, libssh.a and libopenbsd-compat.a to the rest of the world. There are 2 ways:

1. I can contact the owner pf OpenSSH to see if they would accept a patch to support it direclty via make their make install  
2. We can have OpenSSH inherit lib_package class and start exposing manually in a install hook the files we want.

I am also happy with HIBA being a submod of OpenSSH as this is in line with reality

wltu: I think the issues right now is that each sysroot in bitbake is based on recipes, so we can't get reliability get the build folder of another repo. So I think `lib_package` will hit the same issue.

Yeah, seems like a submodule in openssh is the best move and maybe we don't use the official release of openssh for `hiba` until they are available. Build from SRC_URI and such.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: We have now installed the files with an openssh bbappend. @blunderer has pinged the OpenSSH maintainer to install it on their end, but that probably need an official release and may take a while. The bbappend will be needed until then.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Thanks @blunderer's feedback, I have updated the latest recipe changes. It seems like I just need to install the headers + lib*.so for it to work. I thought we needed the object files and was hesitant to do so.

It should be better now.

kraj: its better than before. However, I think it still needs modifications in openssl as I grok. If it does, then get those changes into openssh recipe first, the recipe needs to build with standard openssh provided by oe-core for it to be in any open layers.

wltu: sounds good. Thanks make sense. We can follow up on that before getting this submitted.

If Google/OpenBMC needs it earlier, we will support it within our meta-layer until the openssh changes are ready. and then migrate it here.

Thanks for the help and suggestions.

blunderer: There should be no modification to openssl required since openssl library and includes are installed in the sysroot already and available for HIBA to link against in a standard way.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -17,26 +17,22 @@ PV = "1.0+git${SRCPV}"
...
-  if len(openssh_paths) != 1:
-    # Cannot fail here since this will cause failure on other recipes
-    # if openssh is not installed
-    return ""
-
-  return "--prefix=/usr/ --with-opensshdir={}".format(openssh_paths[0])
-
-EXTRA_OECONF = "${@get_oeconf(d)}"
+EXTRA_OECONF = " \
+  --with-opensshdir=${PKG_CONFIG_SYSROOT_DIR}/usr/include/ssh \
```
kraj, line 22: Please use `${STAGING_INCDIR}/ssh` here

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -17,26 +17,22 @@ PV = "1.0+git${SRCPV}"
...
+  --prefix=/usr/ \
```
kraj, line 23: Use `--prefix=${prefix}`

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -17,26 +17,22 @@ PV = "1.0+git${SRCPV}"
...
+"
 
 # Make sure that MACHINE_FEATURES includes hiba before continuing
 do_configure:prepend(){
-  if ${@bb.utils.contains("MACHINE_FEATURES", "hiba", "false", "true", d)} ; then
-    echo "MACHINE_FEATURES does not support hiba"
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
+    echo "DISTRO_FEATURES does not support hiba. Skip the configuration"
```
kraj, line 29: Consider something like below for diagnostic message 
`Skip building ${PN} because hiba support is disabled, It can be enabled it via DISTRO_FEATURES`

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -17,26 +17,22 @@ PV = "1.0+git${SRCPV}"
...
     exit 1
   fi
 }
+
+PACKAGES =+ "${PN}-tools"
+
+FILES:${PN}-tools = "${sbindir}/hiba-ca.sh ${sbindir}/hiba-grl ${sbindir}/hiba-gen"
+
+RDEPENDS:${PN}-tools = "${PN} bash"
```
kraj, line 38: Are these bash scripts ?  We avoid bash dependencies as much as we can and enable it absolutely when needed.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -0,0 +1,14 @@
+do_install:append () {
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
+    echo "DISTRO_FEATURES does not support hiba. Skip the installations"
+    exit 0
+  fi
```
kraj, line 5: this maybe redundant.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -0,0 +1,14 @@
...
+
+  install -d ${D}/usr/include/ssh
+  install -d ${D}/usr/include/ssh/openbsd-compat
+  install -d ${D}/usr/lib/
```
kraj, line 9: Use bitbake variables instead of hardcoding paths where ever possible.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -0,0 +1,14 @@
...
+  install -m0644 ${S}/*.h ${D}/usr/include/ssh
+  install -m0644 ${S}/libssh* ${D}/usr/lib/
+  install -m0644 ${S}/openbsd-compat/*.h ${D}/usr/include/ssh/openbsd-compat
+  install -m0644 ${S}/openbsd-compat/libopenbsd-compat* ${D}/usr/lib/
```
kraj, line 13: where are these files coming from ?

kraj (changes requested):
also squash changes into a single commit.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
wltu: rm'ed and will be moved to poky recipe once I get write access.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
wltu: Done. And moving to poky recipe

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: I don't think we need that script during runtime? (maybe).

actually, we might need it. Let's see what @blunderer think.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Done.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Done.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Done.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Will move the recipe change to poky for now.

meta-oe/recipes-connectivity/openssh/openssh_%.bbappend
wltu: These are all of the files that's build from https://github.com/openssh/openssh-portable, or do you mean something else?

kraj:
OK I will wait for the openssh changes to be merged before considering this patch.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Poky review in https://git.yoctoproject.org/poky-contrib/commit/?h=wltu/install_openssh_files&id=c8441295669fd11c0fe3f88f943d3afcd5fc172c. PTAL.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
blunderer: We do not need the bash script on the target, this is mostly for the host

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Oops. I updated to https://git.yoctoproject.org/poky-contrib/?h=wltu%2Finstall_internal_openssh_files and cc you guys in the email. I am not sure how to get more attention to the change or if I did anything wrong.

Any tips?

wltu:
> OK I will wait for the openssh changes to be merged before considering this patch.

Updated the change to install an internal version of OpenSSH to get the library required based on the comments in this discussion: https://lists.openembedded.org/g/openembedded-core/topic/94475279#172977

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+
+S = "${WORKDIR}/git"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=7e161abe9a4585310725dd86c28d1ae2"
+
+SRC_URI = "git://github.com/google/hiba;protocol=https;branch=main"
+SRCREV = "98b551e2a0297b1872851bb81b6d3a2b682778ee"
+
+# HIBA Internal OpenSSH Version (Should match the version in meta-oe).
```
blunderer, line 15: This is actually not an issue: HIBA only uses a small part of the libssh which happens to be part of the RFC and thus pretty stable, so we can technically use pretty much any OpenSSH version.

Suggestion:

Internal OpenSSH version for HIBA to link against (the version doesn't have to match the actual installed version).

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+inherit autotools
+
+DEPENDS = "openssl openssh zlib"
+
+EXTRA_OECONF = " \
+  --with-opensshdir=${WORKDIR}/openssh-portable \
+  --prefix=/usr/ \
+"
+
+# Make sure that MACHINE_FEATURES includes hiba before continuing
```
blunderer, line 32: nit: DISTRO_FEATURE

blunderer: Do we actually still need this hack? This was initially used to tune the build of OpenSSH for HIBA, but now everything is self contain, so simply including the HIBA package should be enough.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+do_configure:prepend(){
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
```
blunderer, line 34: And we can get rid of this

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+  git checkout "${OPENSSH_SRCREV}"
+  autoreconf
+  ./configure ${CONFIGUREOPTS}
+  oe_runmake
+  popd
+}
+
+do_install:append() {
+  rm ${D}${sbindir}/hiba-ca.sh
+  rm -rf ${WORKDIR}/openssh-portable
```
blunderer, line 51: Hum, that doesn't look right: a BUILD step for HIBA will now fail after the install

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
+    echo "Skip building ${PN} because hiba support is disabled, It can be enabled it via DISTRO_FEATURES"
+    exit 1
+  fi
+
+  pushd "${WORKDIR}"
+  git clone "${OPENSSH_SRC_URI}"
+  cd openssh-portable
+  git checkout "${OPENSSH_SRCREV}"
+  autoreconf
```
blunderer, line 43: Shouldn't this go better in a do_fetch:append() ?

blunderer: And I believe this can even be done in a more idiomatic way:

LIC_FILES_CHKSUM_hiba = "file://${S}/LICENSE;md5=7e161abe9a4585310725dd86c28d1ae2"
LIC_FILES_CHKSUM_openssh = "file://${S}/LICENSE;md5=7e161abe9a4585310725dd86c28d1ae2"
SRC_URI = "\
    git://github.com/google/hiba;protocol=https;branch=main;destsuffix=git/hiba;name=hiba \
    git://github.com/openssh/openssh-portable.git;protocol=https;branch=master;destsuffix=git/openssh-portable;name=openssh \
"
SRCREV_hiba = "3199239c6a0a89dbd8f155204f44798c78195ab0"
SRCREV_openssh = "0ffb46f2ee2ffcc4daf45ee679e484da8fcf338c"

SRC_URI[hiba.sha256sum] = "79154c8cc144905d552540ef7b395289a029a43ac262bb913b99f475d6da1bec"
SRC_URI[openssh.sha256sum] = "f7a2969530cc025dbbe17b40c1413693e5b57cee04a194ba932b6b2465860bba"

SRCREV_FORMAT="hiba_openssh"
PV = "1.0+git${SRCPV}"

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+
+# Make sure that MACHINE_FEATURES includes hiba before continuing
+do_configure:prepend(){
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
+    echo "Skip building ${PN} because hiba support is disabled, It can be enabled it via DISTRO_FEATURES"
+    exit 1
+  fi
+
+  pushd "${WORKDIR}"
+  git clone "${OPENSSH_SRC_URI}"
```
kraj, line 40: move this cloning to SRC_URI so the fetcher can take care of cloning and ensuring we capture the state correctly for further processing.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,52 @@
...
+"
+
+# Make sure that MACHINE_FEATURES includes hiba before continuing
+do_configure:prepend(){
+  if ${@bb.utils.contains("DISTRO_FEATURES", "hiba", "false", "true", d)} ; then
+    echo "Skip building ${PN} because hiba support is disabled, It can be enabled it via DISTRO_FEATURES"
+    exit 1
+  fi
+
+  pushd "${WORKDIR}"
```
kraj, line 39: do not use `pushd/popd` they may not work on all shells, we do support non-bash shells like dash as default shells on build machines.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
wltu: Updated. PTAL.

wltu: actually, nevermind. Still some failures.

wltu:
friendly ping.

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,50 @@
...
+  cd "${WORKDIR}/git/openssh-portable"
+  autoreconf
+  ./configure ${CONFIGUREOPTS}
+  cd "${S}"
+}
+
+do_compile_openssh() {
+  cd "${WORKDIR}/git/openssh-portable"
+  oe_runmake
+  cd "${S}"
```
blunderer, line 46: Here and above, this is probably not needed

meta-oe/recipes-connectivity/hiba/hiba_git.bb
blunderer: Is it working now? I still do not see SRC_URI checksums for hib and OpenSSH. Is it expected?

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,50 @@
...
+S = "${WORKDIR}/git/hiba"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=7e161abe9a4585310725dd86c28d1ae2"
+
+PV = "git${SRCPV}"
+
+inherit autotools
+
+DEPENDS = "openssl openssh zlib"
+RDEPENDS:${PN} = " bash"
```
blunderer, line 16: We do not need the hiba-ca.sh on the target host, so 'bash' is not needed in RDEPENDS.

blunderer (changes requested):

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,50 @@
...
+
+# Internal OpenSSH version for HIBA to link against (the version doesn't have
+#   to match the actual installed version).
+# Discussion in
+# https://lists.openembedded.org/g/openembedded-core/topic/94475279#172977
+SRCREV_openssh = "15a01cf15f396f87c6d221c5a6af98331c818962"
+SRCREV_FORMAT = "hiba"
+
+EXTRA_OECONF = " \
+  --with-opensshdir=${WORKDIR}/git/openssh-portable \
```
kraj, line 32: Can this be made relative ? absolute paths are troublesome for reproducibility

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,50 @@
...
+}
+
+do_compile_openssh() {
+  cd "${WORKDIR}/git/openssh-portable"
+  oe_runmake
+  cd "${S}"
+}
+
+addtask compile_openssh before do_configure after do_configure_openssh
+addtask configure_openssh before do_compile_openssh after do_unpack
```
kraj, line 50: do_configure should be added after `do_patch` instead of `do_unpack` that way you can apply patches as needed to openssh in future

kraj (changes requested):

meta-oe/recipes-connectivity/hiba/hiba_git.bb
```diff
@@ -0,0 +1,50 @@
...
+inherit autotools
+
+DEPENDS = "openssl openssh zlib"
+RDEPENDS:${PN} = " bash"
+
+SRC_URI = " \
+  git://github.com/google/hiba;protocol=https;branch=main;destsuffix=git/hiba;name=hiba \
+  git://github.com/openssh/openssh-portable.git;protocol=https;branch=master;destsuffix=git/openssh-portable;name=openssh \
+"
+SRCREV_hiba = "cae606dafc5692240ac49441fd57653d7aade99b"
```
blunderer, line 22: I added some changes to HIBA recently.

The new head is at 3a6beb475ebd4dff3223327039738b30240f0c43

PR #598, opened 2022-08-29
unifex: remove recipe

williamspatrick:
I don't think anyone else is using this since nobody has requested
an update from the original SRCREV I pushed up originally.  The
original authors of libunifex have moved in another direction so I've
decided not to use it in my downstream project(s) also.

Anyone wanting to use something like libunifex should probably look
to https://github.com/brycelelbach/wg21_p2300_std_execution , which
is where the proposals for future C++ standards in this space are
being developed.

Signed-off-by: Patrick Williams <patrick@stwcx.xyz>


kraj:
applied with f846d3934d

PR #602, opened 2022-09-06
dbus-broker: update to v32

bluca:
https://github.com/bus1/dbus-broker/releases/tag/v32


kraj:
Fixed with 5a02e200ff

PR #612, opened 2022-09-21
rrdtool: make graph dependencies optional

abeltranmsft:
Make graph dependencies (pango and cairo) optional through PACKAGECONFIG. Enable them by default


kraj:
merged with 9c136fcd9e

PR #613, opened 2022-09-24
libwebsockets: add optional support for sd-event loop

lrusak:
Hello!

I'm sending a pull request to add support for building the libwebsockets-evlib-sd plugin for libwebsockets.

This allows using the sd_events based loop.

I am using this in a custom app and thought I would contribute the change.

### DIFF
--- a/meta-oe/recipes-connectivity/libwebsockets/libwebsockets_4.3.2.bb
+++ b/meta-oe/recipes-connectivity/libwebsockets/libwebsockets_4.3.2.bb
@@ -26,6 +26,11 @@ PACKAGECONFIG[ssl] = "-DLWS_WITH_SSL=ON,-DLWS_WITH_SSL=OFF,openssl"
 PACKAGECONFIG[static] = "-DLWS_WITH_STATIC=ON,-DLWS_WITH_STATIC=OFF -DLWS_LINK_TESTAPPS_DYNAMIC=ON,"
 PACKAGECONFIG[systemd] = "-DLWS_WITH_SDEVENT=ON,-DLWS_WITH_SDEVENT=OFF,systemd"
 
+python __anonymous() {
+  if bb.utils.contains('PACKAGECONFIG', 'systemd', True, False, d) and not bb.utils.contains('DISTRO_FEATURES', 'systemd', True, False, d):
+    bb.fatal("PACKAGECONFIG contains systemd but DISTRO_FEATURES doesn't")
+}
+
 EXTRA_OECMAKE += " \
     -DLIB_SUFFIX=${@d.getVar('baselib').replace('lib', '')} \
 "


meta-oe/recipes-connectivity/libwebsockets/libwebsockets_4.3.2.bb
```diff
@@ -24,21 +24,24 @@ PACKAGECONFIG[libuv] = "-DLWS_WITH_LIBUV=ON,-DLWS_WITH_LIBUV=OFF,libuv"
 PACKAGECONFIG[server] = "-DLWS_WITHOUT_SERVER=OFF,-DLWS_WITHOUT_SERVER=ON,"
 PACKAGECONFIG[ssl] = "-DLWS_WITH_SSL=ON,-DLWS_WITH_SSL=OFF,openssl"
 PACKAGECONFIG[static] = "-DLWS_WITH_STATIC=ON,-DLWS_WITH_STATIC=OFF -DLWS_LINK_TESTAPPS_DYNAMIC=ON,"
+PACKAGECONFIG[systemd] = "-DLWS_WITH_SDEVENT=ON,-DLWS_WITH_SDEVENT=OFF,systemd"
```
kraj, line 27: this `PACKAGECONFIG` needs to be enabled as well when systemd is in `DISTRO_FEATURES`

kraj (changes requested):

meta-oe/recipes-connectivity/libwebsockets/libwebsockets_4.3.2.bb
lrusak: I thought of that but ultimately I left it out because it really is an optional feature and shouldn't be automatically enabled just because a distro enables systemd.

kraj: OK, thats good, however it will be nearly never used by folks since they would assume its always tied to distro features and secondly, since it has dependency on systemd somehow we need to also ensure that this packageconfig when enabled without systemd gives an error if we do not tie it to distro feature.

lrusak: I've added a fixup to check this condition. Please check if that is acceptable or if you would like the error message changed.

kraj: Please submit that in a new pull request since original patch is already applied

lrusak: > Please submit that in a new pull request since original patch is already applied

I noticed that after I pushed, sorry 😞.

I updated the PR to contain just the new commit, do you want to just pick it or do you want me to open an new PR, or is this one still ok?

kraj:
fixed with f3fa2a5d5c02d8937ee5c866b827a9553c9dd351

PR #615, opened 2022-09-27
tio: added tio version 2.0

onkelpit:
tio: added tio version 2.0
From the Github repo:
tio - a simple serial device I/O tool

tio is a simple serial device tool which features a straightforward command-line and configuration file interface to easily connect to serial TTY devices for basic I/O operations.

### DIFF
--- a/meta-oe/recipes-support/tio/tio_2.0.bb
+++ b/meta-oe/recipes-support/tio/tio_2.0.bb
@@ -1,0 +1,19 @@
+SUMMARY = "tio - a simple serial device I/O tool"
+DESCRIPTION = "tio is a simple serial device tool which features a \
+    straightforward command-line and configuration file interface to easily \
+    connect to serial TTY devices for basic I/O operations."
+
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0e1a95b7892d3015ecd6d0016f601f2c"
+
+SRC_URI = "git://github.com/tio/tio;protocol=https;nobranch=1"
+SRCREV = "6618642acf28fec6d3e70ed75b50d4ce138ea08a"
+
+S = "${WORKDIR}/git"
+
+inherit meson pkgconfig
+
+DEPENDS += " libinih"
+RDEPENDS:${PN} += " libinih"
+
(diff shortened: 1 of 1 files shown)


meta-oe/recipes-support/tio/tio.inc
```diff
@@ -0,0 +1,16 @@
...
+
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0e1a95b7892d3015ecd6d0016f601f2c"
+
+S = "${WORKDIR}/git"
+
+inherit meson pkgconfig
+
+DEPENDS += " libinih"
+RDEPENDS_${PN} += " libinih"
```
kraj, line 14: this should be `RDEPENDS:${PN}`

meta-oe/recipes-support/tio/tio_1.47.bb
```diff
@@ -0,0 +1,3 @@
+require tio.inc
```
kraj, line 1: I think we dont want the 1.xy version unless you convince otherwise.

meta-oe/recipes-support/tio/tio_2.0.bb
```diff
@@ -0,0 +1,3 @@
+require tio.inc
+SRC_URI = "git://github.com/tio/tio;protocol=https;nobranch=1"
```
kraj, line 2: why `nobranch`

kraj (changes requested):

meta-oe/recipes-support/tio/tio.inc
onkelpit: Done.

kraj:
please squash all commits into a single commit and refresh the PR

meta-oe/recipes-support/tio/tio_2.0.bb
onkelpit: So that the recipe still works, if someone renames the default branch to main or something else. If it's not wanted, i could change that too.

kraj:
there is no need for a `.inc` file, please merge all content into `.bb` file and upload again

kraj:
fixed with bac3fe527a2dca3b2fe5612535589a31ce514d5d

PR #623, opened 2022-10-12
mctp: install the .target files

drakedog2008:
Need the targets file to enable the mctpd.service on systemd.

Signed-off-by: Hao Jiang <jianghao@google.com>
Change-Id: I8d48d3767760dc1f34ae7e1266600d350ac93281


kraj:
applied with 6e2cfa4113

PR #624, opened 2022-10-14
abseil-cpp: Upgrade to head on 2022-10-14

UlysseM:
Our team wants to use absl/log which was recently added to abseil.

### DIFF
--- a/meta-oe/recipes-devtools/abseil-cpp/abseil-cpp/abseil-ppc-fixes.patch
+++ b/meta-oe/recipes-devtools/abseil-cpp/abseil-cpp/abseil-ppc-fixes.patch
@@ -20,7 +20,7 @@ Signed-off-by: Khem Raj <raj.khem@gmail.com>
  #ifdef __GLIBC__
  #include <sys/platform/ppc.h>
  #elif defined(__FreeBSD__)
-@@ -53,7 +53,7 @@ double UnscaledCycleClock::Frequency() {
+@@ -58,7 +58,7 @@ double UnscaledCycleClock::Frequency() {
    return base_internal::NominalCPUFrequency();
  }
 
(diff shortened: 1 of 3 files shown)


kraj:
Seeing patching errors. How was this tested ?

https://autobuilder.yoctoproject.org/typhoon/#/builders/88/builds/2102

UlysseM:
Hi Raj, I'm pretty new to this repo and thought there was a CI up, so I did not run any test.

That being said, from your build, I noticed that the patch files were outdated, so I fixed those.

kraj:
I am also seeing failure building grpc-native on ubuntu-18.04 ( which is a supported host still ) see

https://errors.yoctoproject.org/Errors/Details/673140/

Perhaps abseil library has moved beyond the standard that gcc is supporting in ubuntu 18.04

UlysseM:
According to https://github.com/abseil/abseil-cpp/commit/97ab3dcfd6490434202e4ab00b2eaba9449e42a1, abseil moved to have a minimum check of c++14 / gcc 5.

kraj:
> According to [abseil/abseil-cpp@97ab3dc](https://github.com/abseil/abseil-cpp/commit/97ab3dcfd6490434202e4ab00b2eaba9449e42a1), abseil moved to have a minimum check of c++14 / gcc 5.
> 
> As a side note, the current LTS version we're using (20220623.1) mentioned that this would be the last to support cpp11.

perhaps we need to fix grpc-native to use `-std=c++14` in CXXFLAGS at least on ubuntu 18.04 hosts.

UlysseM:
I'm not super familiar with this build pipeline and I don't know how to run your CI, so hopefully my last update does what you suggested.

kraj:
> I'm not super familiar with this build pipeline and I don't know how to run your CI, so hopefully my last update does what you suggested.

We dont have github actions triggered CI, we manually stage patches into a staging branch `master-next` and then run a CI on it using yocto autobuilders https://autobuilder.yoctoproject.org/typhoon/#/builders/88

Then some downstream distros do their own CI which includes meta-openembedded e.g. https://github.com/YoeDistro/meta-openembedded/actions

but this is only doing subset of packages. Building all the recipes takes a lot of time. So we do not have either compute or man power to do that

kraj:
I have upgraded grpc to 1.50.x which also now requires `c++14` ( since 1.47 onwards ) so it fixed the problem nicely.

kraj:
fixed with 3ab22400d1

PR #625, opened 2022-10-18
[Fix] Add condition for libusbgx-examples

homalozoa:
* Add @bb.utils.contains for PACKAGE_BEFORE_PN and INITSCRIPT_PACKAGES to make configs in PACKAGECONFIG work.

Signed-off-by: homalozoa <nx.tardis@gmail.com>


kraj:
applied with 36f92899a6

PR #629, opened 2022-10-25
ipmitool: add default iana enterprise numbers database

chenxy1988:
When using default configuration without add IANA_ENTERPRISE_NUMBERS and SRC_URI in a bbappend, the ipmitool would report an error as below:

"
root@intel-x86-64:~# ipmitool -I open bmc getenables IANA PEN registry open failed: No such file or directory "

Add a default iana enterprise numbers database to avoid the error happen if using ipmitool default bb configuration.

Signed-off-by: Xiangyu Chen <xiangyu.chen@windriver.com>


meta-oe/recipes-kernel/ipmitool/ipmitool_1.8.19.bb
```diff
@@ -27,10 +27,12 @@ SRC_URI = "git://github.com/ipmitool/ipmitool;protocol=https;branch=master \
            file://0001-ipmi_fru.c-Provide-missing-function-declarations.patch \
            file://0001-configure-Remove-the-logic-to-download-IANA-PEN-data.patch \
            "
-IANA_ENTERPRISE_NUMBERS ?= ""
 
-# Add these via bbappend if this database is needed by the system
-#IANA_ENTERPRISE_NUMBERS ?= "http://www.iana.org/assignments/enterprise-numbers;name=iana-enterprise-numbers;downloadfilename=iana-enterprise-numbers"
+# Default IANA enterprise database version is 2022-10-24
+IANA_ENTERPRISE_NUMBERS ?= "file://iana-enterprise-numbers.tar.xz"
```
kraj, line 32: Can we host this tarball somewhere and download it during build ? git does not do well with binary blobs particularly. I am sure we will bump revisions of this as well in future.

kraj (changes requested):

meta-oe/recipes-kernel/ipmitool/ipmitool_1.8.19.bb
chenxy1988: thanks for your reply ;p
Indeed, it's a bit hard to manage with git, the information in tarball from the IANA site, it's a pure text file to record vender/enterprise database the size is almost 4.4MB,  it seems a bit large so I created a tarball.

chenxy1988: I think we can keep a default enterprise database to make sure ipmitool is working normal, if a user wants to use the latest version database, they can add a bbappend file to turn on "IANA_ENTERPRISE_NUMBERS = "http://www.iana.org/assignments/enterprise-numbers;name=iana-enterprise-numbers;downloadfilename=iana-enterprise-numbers"" in their own layer.

chenxy1988: > Can we host this tarball somewhere and download it during build ? git does not do well with binary blobs particularly. I am sure we will bump revisions of this as well in future.

thanks for your reply ;p
Indeed, it's a bit hard to manage with git, the information in tarball from the IANA site, it's a pure text file to record vender/enterprise database the size is almost 4.4MB, it seems a bit large so I created a tarball.

I think we can keep a default enterprise database to make sure ipmitool is working normal, if a user wants to use the latest version database, they can add a bbappend file to turn on "IANA_ENTERPRISE_NUMBERS = "http://www.iana.org/assignments/enterprise-numbers;name=iana-enterprise-numbers;downloadfilename=iana-enterprise-numbers"" in their own layer.

thanks,
Br,
Xiangyu

kraj:
it would be fine to put it somewhere on https/ftp site.

chenxy1988:
> it would be fine to put it somewhere on https/ftp site.

After investigating the enterprise numbers database text file, I think we can find a compromise propose. 
The database recorded enterprise/vendor private enterprise Codes for SMI Network Management, it keeps updated when has new enterprise join. This database has a fixed format and has a reserve(default) code as below:

Decimal
| Organization
| | Contact
| | | Email
| | | |
0
  Reserved
    Internet Assigned Numbers Authority
      iana&iana.org
      
We can create a text file just contains reserved code instead of the full data in a compressed binary blob for ipmitool normal working when using yocto generic build without any change, if a user needs full SMI network Management feature, they can enable the IANA_ENTERPRISE_NUMBERS and SRC_URI in their own layer.


Thanks,

Br,
Xiangyu

PR #631, opened 2022-10-28
android-tools-conf-configfs: Allow handling two or more UDC controllers

bhupesh-sharma:
Several Qualcomm platforms support setting two or more (or even all) UDC controllers in 'peripheral' mode. Modify the android-tools-conf-configfs scriptware to handle such cases.

Signed-off-by: Bhupesh Sharma <bhupesh.sharma@linaro.org>


kraj:
should be in with e87a9506f0494e6d7c420adeb77dff5b25965170

PR #633, opened 2022-11-03
Update libcrypt-openssl-rsa-perl - v0.32 unavailable

Dexus:
This PR update to the last version that is available to download.

0.32 is not only available anymore.


kraj:
fixed with 3459e017b8 on kirkstone branch

PR #637, opened 2022-11-30
add native BBCLASSEXTEND to libbpf

krsh:
To build `pahole-native` we need `libbpf-native`.


kraj:
applied with a28b7fdbf4

PR #644, opened 2022-12-11
python3-twisted: Add python3-asyncio to RDEPENDS

Hains:
To fix crash due to missing module:

from twisted.internet import defer
File "/usr/lib/python3.11/site-packages/twisted/internet/defer.py", line 14, in <module> from asyncio import AbstractEventLoop, Future, iscoroutine ModuleNotFoundError: No module named 'asyncio'

Signed-off-by: Hains van den Bosch <hainsvdbosch@ziggo.nl>


kraj:
applied with 78386ad01d

PR #649, opened 2023-01-12
Make tvheadend recipe compatible with the latest tvheadend version

pbasista:
Update the tvheadend recipe and the included patch so that it is compatible with the latest upstream version of tvheadend.


akuster:
now in master.

https://git.openembedded.org/meta-openembedded/commit/meta-multimedia/recipes-dvb/tvheadend?id=babfc021766d32578c5ed136897d22cef5abdd24

edisionnano:
It's been 3 years since TVHeadend was updated. Would you be able to bump it again?

PR #652, opened 2023-01-23
uutils-coreutils: upgrade 0.0.16 -> 0.0.17

Ecordonnier:


kraj:
applied with eec98799a2

PR #655, opened 2023-01-30
cfengine: upgrade to 3.21.0

craigcomstock:
From 3.15.0

Our ticket for this work:
https://tracker.mender.io/browse/CFE-3650


craigcomstock:
probably you don't want a patch to hardknott... will re-work

PR #661, opened 2023-03-06
fluentbit: change of download name

onkelpit:
To prevent naming collisions when BB_GENERATE_MIRROR_TARBALLS is used, the packagename is used for the downloaded file.

Otherwise it would just be source-${PV}.tar.gz


kraj:
applied with 827bce8217, @akuster we might want it backported to older releases as applicable

shr-project:
The change is good, but don't see how it would conflict with BB_GENERATE_MIRROR_TARBALLS which always start with git2_ (or some other scm) prefixes.

FWIW: I was checking how many source-* files I have on my PREMIRROR and found another gem:
source-code-pro-2.030R-ro-1.050R-it.zip but google confirmed it's kind of valid in this case as it's from:
https://cgit.openembedded.org/meta-openembedded/plain/meta-oe/recipes-graphics/ttf-fonts/source-code-pro-fonts_2.030_1.050.bb

PR #664, opened 2023-03-14
Add android-tools version 10

Ecordonnier:
As discussed in https://github.com/kraj/meta-clang/pull/756#issuecomment-1461969719 , I am moving this recipe from meta-clang to meta-openembedded.

Please don't merge it yet, because I still want to test it on a generic raspberry-pi hardware (I tested it on custom hardware). However I think opening the PR now makes sense in order to get early feedback.

Follow-up work: update to a more recent version (requires adding another debian repository) and also evaluate whether https://github.com/nmeum/android-tools.git can be used instead of the debian sources


kraj:
Please move these recipes to `dynamic-layers/selinux/android-tools/` folder and add selinux to dynamic layers in `layer.conf`

Ecordonnier:
> Please move these recipes to `dynamic-layers/selinux/android-tools/` folder and add selinux to dynamic layers in `layer.conf`

What do you mean? The recipe is already in `dynamic-layers/selinux/recipes-devtools/android-tools/` and selinux is already added as dynamic layer to layer.conf.

kraj:
> > Please move these recipes to `dynamic-layers/selinux/android-tools/` folder and add selinux to dynamic layers in `layer.conf`
> 
> What do you mean? The new recipe is already in `dynamic-layers/selinux/recipes-devtools/android-tools/` and selinux is already added as dynamic layer to layer.conf.

I confused myself, there is 5.1.1 recipe still in ./meta-openembedded/meta-oe/recipes-devtools/android-tools/
this should be removed I think

Ecordonnier:
> > > Please move these recipes to `dynamic-layers/selinux/android-tools/` folder and add selinux to dynamic layers in `layer.conf`
> > 
> > 
> > What do you mean? The new recipe is already in `dynamic-layers/selinux/recipes-devtools/android-tools/` and selinux is already added as dynamic layer to layer.conf.
> 
> I confused myself, there is 5.1.1 recipe still in ./meta-openembedded/meta-oe/recipes-devtools/android-tools/ this should be removed I think

Yes, I was not sure how to handle this because the version 10 depends on selinux using DEPENDS while the version 5 builds its own version of selinux and is thus formally dependent on selinux. I guess there could be corner-cases where people don't have selinux as a DISTRO_FEATURE and their setup would break once the version 5 is removed. If you're OK with that I can add a commit removing the version 5.

kraj:
yeah perhaps leave both in.

kraj:
its now in meta-oe. Please send a PR to remove it from meta-clang as well.

Ecordonnier:
Thanks, I sent a PR for removal from meta-clang at https://github.com/kraj/meta-clang/pull/761

PR #665, opened 2023-03-15
android-tools: fix TMPDIR

Ecordonnier:
Remove the TMPDIR environment variable from adb which is used by commands like mktemp (set to /data/local/tmp in android but vanilla linux uses /tmp per default).


kraj:
applied with b71b1eedb4

PR #667, opened 2023-03-22
android-tools: update to 29.0.6.r14

Ecordonnier:
Note: the adbd makefile can't be upstreamed to debian, because there is already an open PR upstream:
https://salsa.debian.org/android-tools-team/android-platform-tools/-/merge_requests/4/diffs


Ecordonnier:
@kraj I am in the process of creating an account on salsa, so that I can upstream some of those patches like you requested in the previous PR. At the moment I have no salsa account, though (as far as I understand those are validated manually).

kraj:
> @kraj I am in the process of creating an account on salsa, so that I can upstream some of those patches like you requested in the previous PR. At the moment I have no salsa account, though (as far as I understand those are validated manually).

ok, let me know if progress.

kraj:
applied with 3add35fa2d

PR #670, opened 2023-04-03
uutils coreutils 0.0.18

Ecordonnier:


kraj:
fails to build on musl/qemuarm64

```
| error[E0432]: unresolved imports `uucore::utmpx`, `uucore::utmpx`                                                                                                    
|   --> src/uu/pinky/src/pinky.rs:13:13                                                                                                                                
|    |                                                                                                                                                                 
| 13 | use uucore::utmpx::{self, time, Utmpx};                                                                                                                         
|    |             ^^^^^   ^^^^ no `utmpx` in the root                                                                                                                 
|    |             |                                                                                                                                                   
|    |             could not find `utmpx` in `uucore`                                                                                                                  
|                                                                                                                                                                      
| error[E0432]: unresolved imports `uucore::utmpx`, `uucore::utmpx`                                                                                                    
|   --> src/uu/users/src/users.rs:17:13                                                                                                                                
|    |                                                                                                                                                                 
| 17 | use uucore::utmpx::{self, Utmpx};                                                                                                                               
|    |             ^^^^^   ^^^^ no `utmpx` in the root                                                                                                                 
|    |             |                                                                                                                                                   
|    |             could not find `utmpx` in `uucore`                                                                                                                  
|                                                                                                                                                                      
| error[E0432]: unresolved imports `uucore::utmpx`, `uucore::utmpx`                                                                                                    
|   --> src/uu/who/src/who.rs:13:13                                                                                                                                    
|    |                                                                                                                                                                 
| 13 | use uucore::utmpx::{self, time, Utmpx};                                                                                                                         
|    |             ^^^^^   ^^^^ no `utmpx` in the root                                                                                                                 
|    |             |                                                                                                                                                   
|    |             could not find `utmpx` in `uucore`                                                                                                                  
|                                                                                                                                                                      
```

Ecordonnier:
I investigated the issue, and the root-cause is that https://github.com/uutils/coreutils/blob/90609044ebc8a6802bb2aef05c57600b4de5af1e/src/uucore/src/lib/features.rs#L42 disables utmpx when building uucore for the musl target, as a work-around for https://github.com/uutils/coreutils/issues/1361 
The reason this was done is that the libc crate has not implemented functions like "endutxent" for musl (this file needs to be modified: https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/musl/mod.rs).

Is there a way to disable the build of uutils-coreutils for musl in yocto, so that the PR is unblocked? Or is the implementation of uutils-coreutils for musl a pre-requisite?

kraj:
> 

its fine to disable it for musl. Perhaps do 

`COMPATIBLE_HOST:libc-musl = "null"`

although it will be good to inform upstream rust libc about this problem and see if it can be fixed and tracked.

Ecordonnier:
I added the line you recommended to disable musl. I also created a ticket at libc to track the issue: https://github.com/rust-lang/libc/issues/3190

Is there some CI which tries to build all recipes with musl, and will fail because of the following error when trying to build with musl?

`ERROR: Nothing PROVIDES 'uutils-coreutils'
uutils-coreutils was skipped: incompatible with host x86_64-poky-linux-musl (not in COMPATIBLE_HOST)
`

kraj:
applied as of 6b197f009d

PR #671, opened 2023-04-03
redirect unwanted error message in nginx install

lumpyh:
if we run opkg install nginx on our system (without systemd) we end up getting the following message in the install process

$ opkg install nginx_1.20.1-r0_core2-64.ipk 
...
//var/lib/opkg/info/nginx.postinst: line 3: type: systemd-tmpfiles: not found

this confused some of my coworkers.
as installation also finishes correctly without sytemd-tmpfiles and not having systemd-tempfiles is not really a problem, I think we should redirect the message also to /dev/NULL


kraj:
applied with d13bea0481

PR #673, opened 2023-04-05
Add ptest for python recipes (lorem, path) and double-conversion

ycongal-smile:


kraj:
applied as of 1dc0260309

ycongal-smile:
Thanks!

PR #674, opened 2023-04-06
meta-oe: recipes-extended: bitwise: Add ptest

ycongal-smile:


kraj:
applied with f19dac11ff

PR #677, opened 2023-04-11
qrencode: add PACKAGECONFIG for command line tools, enable native build

rohieb:


kraj:
applied with 5a50070964

PR #682, opened 2023-04-27
meta-oe: recipes-devtool: openocd: update version and add interfaces

strumtrar:
Update OpenOCD to the newest release, update Jimtcl URL and enable two new interfaces.


kraj:
applied with a6738ea35b

PR #684, opened 2023-05-08
serial: add recipe for version 1.2.1

mtudan:
Serial is cross-platform library for interfacing with RS-232 serial like ports.

### DIFF
--- a/meta-oe/recipes-support/serial/serial_1.2.1.bb
+++ b/meta-oe/recipes-support/serial/serial_1.2.1.bb
@@ -1,0 +1,28 @@
+SUMMARY = "Cross-platform library for interfacing with rs-232 serial like ports"
+HOMEPAGE = "http://wjwwood.io/serial/"
+LICENSE = "MIT"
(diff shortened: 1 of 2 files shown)


kraj:
Fails to fetch see

https://errors.yoctoproject.org/Errors/Details/703595/

meta-oe/recipes-support/serial/serial_1.2.1.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "Cross-platform library for interfacing with rs-232 serial like ports"
+HOMEPAGE = "http://wjwwood.io/serial/"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://README.md;beginline=53;endline=62;md5=049c68d559533f90250404746e6a1045"
+
+SRC_URI = " \
+    git://github.com/wjwwood/serial.git;nobranch=1;tag=${PV} \
```
kraj, line 7: We should not use `tags` in SRC_URI since they are considered floating, please use SRCREV

kraj (changes requested):

kraj:
applied with 77b6df94c1

PR #687, opened 2023-05-09
open62541: allow overriding encryption providers and disable warnings as errors

JohannesKauffmann:


kraj:
applied with bd7cd3bc6f

PR #690, opened 2023-05-11
paho-mqtt-c: Improve performance and disable tests

ddassie-texa:
The paho-mqtt-c library builds with tests and tracing functionality enabled by default, these patches disable both of the settings improving runtime and build performance.


kraj:
applied with bf26b3e5ed

PR #691, opened 2023-05-12
 etcd: systemd unit support for clustering

ojayanth:
Refer  etcd documentation https://etcd.io/docs/v3.5/op-guide/clustering/
for etcd based cluster design details.

Added  required systemd unit and configuration files to start a new cluster and restart existing etcd cluster.


kraj:
applied with a179653ce2

PR #692, opened 2023-05-16
hoth: Add recipe for libhoth

wltu:
libhoth is the google specific usb protocol implementation which is required by hothd.

### DIFF
--- a/meta-oe/recipes-extended/hoth/libhoth_git.bb
+++ b/meta-oe/recipes-extended/hoth/libhoth_git.bb
@@ -1,0 +1,16 @@
+
+SUMMARY = "Google Hoth USB library"
+DESCRIPTION = "A library implements Google USB protocol to communication with Hoth device"
+LICENSE = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


wltu:
Not sure if we should even have this in meta-oe, but opening this to remove it from openbmc. https://gerrit.openbmc.org/c/openbmc/openbmc/+/63424

kraj:
> Not sure if we should even have this in meta-oe, but opening this to remove it from openbmc. https://gerrit.openbmc.org/c/openbmc/openbmc/+/63424

this seems a simple recipe, however, are there wider userbase you are expecting for this library than openBMC ?
some other layer needing it ? That might be a good reason.

wltu:
@cjevans-google, any thoughts on this? If this will only be in gBMC then we should move it to https://gbmc.googlesource.com/meta-gbmc-staging/ instead

akuster:
This recipe is in meta-security.

akuster:
Google submitted the recipe:
https://git.yoctoproject.org/meta-security/commit/meta-tpm?id=71199365ff5159f7b2084741b785604bfb403d11

wltu:
oh thanks. I wasn't  aware of that. (and that person is on my team. oops.

I will close this and bump that source rev instead. :)

PR #693, opened 2023-05-18
python3-aioredis: add recipe

ricoheller:
This library provides an interface to Redis based on asyncio.


kraj:
applied with 67b7a44962

PR #694, opened 2023-05-18
python3-dbus-fast: add recipe

ricoheller:
This library provides a Python 3 interface to DBus.


kraj:
applied with 67b7a44962

PR #697, opened 2023-05-23
python3-bleak: add recipe

ricoheller:
Bleak is a GATT client software, capable of connecting to BLE devices acting as GATT servers.


kraj:
applied with 761e5ffffd

PR #699, opened 2023-05-28
bdwgc: upgrade 8.2.2 -> 8.2.4

ivmai:
Also, HOMEPAGE is updated to use https.


kraj:
applied with 902725991f

PR #700, opened 2023-05-31
open62541: update to v1.3.6

JohannesKauffmann:
Since v1.3.4, support for OpenSSL 3.0 has been added.


kraj:
applied with dffc4afc53

PR #702, opened 2023-06-05
open62541: build optimized binary

JohannesKauffmann:
Without any build type specified, open62541 defaults to "Debug".


kraj:
applied with 5bbfb95eed

PR #705, opened 2023-06-07
meta-python: Add stopit

azaharia-wr:
stopit is a library to raise asynchronous exceptions in other threads, control the timeout of blocks or callables with two context managers and two decorators.

The license isn't shipped in the pypi package, so it was extracted from the github repo at [1].

[1] https://github.com/glenfant/stopit/blob/master/LICENSE


kraj:
applied with 586a4bfbac

PR #706, opened 2023-06-07
uutils-coreutils: upgrade 0.0.18 -> 0.0.19

Ecordonnier:
Note: the removal of "name=XXX" in the crates.inc file is normal and due to a change in cargo-update-recipe-crates.bbclass in poky.


kraj:
applied with 8d38aec39c

PR #709, opened 2023-06-12
python3-stopit: add missing run-time dependencies

azaharia-wr:
The stopit package needs setuptools python module so add it to RDEPENDS.

This fixes:
    File "/usr/lib64/python3.9/site-packages/stopit/__init__.py", line 10, in <module>
	import pkg_resources
    ModuleNotFoundError: No module named 'pkg_resources'


kraj:
applied with 812f091414

PR #710, opened 2023-06-13
Add more open62541 PACKAGECONFIG options

JohannesKauffmann:
This shouldn't change the defaults, but adds more configurable knobs.


kraj:
applied at f8f47d5732

PR #716, opened 2023-07-17
nginx: kirkstone: upgrade to 1.24.0 release

hoinmic:
According to http://nginx.org/en/CHANGES nginx supports the openssl 3.x component only from version 1.21.2. In Kirstone openssl 3.x is included but all provided versions of nginx are older, so there is currently an incompatibility. With this patch this incompatibility get removed.


kraj:
@akuster I think the reasoning for including this into kirkstone seems legit to me. We have 1.21.1 and if it does not work with version of openssl in kirkstone then we have an issue to address.

akuster:
in kirkstone-nut for processing.

kraj:
applied with 8a75c61cce

PR #721, opened 2023-07-22
mstpd: Add initial recipe for mstpd

wltu:
This adds the mstpd package, a daemon implementing RSTP and MSTP protocols in place of the kernel STP implementation.


kraj:
accepted into master with 4c201ede93

PR #722, opened 2023-07-24
android-tools: fix QA warning about buildpaths

chris2511:
Replace all filenames in the object with "--file-prefix-map" which is lost when resetting CFLAGS.

### DIFF
--- a/meta-oe/recipes-devtools/android-tools/android-tools_5.1.1.r37.bb
+++ b/meta-oe/recipes-devtools/android-tools/android-tools_5.1.1.r37.bb
@@ -92,7 +92,7 @@ do_compile() {
 
     # Setting both variables below causing our makefiles to not work with
     # implicit make rules
-    unset CFLAGS
+    CFLAGS="-ffile-prefix-map=${S}=/usr/src/debug/${PN}/${EXTENDPE}${PV}-${PR}"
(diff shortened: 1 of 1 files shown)


meta-oe/recipes-devtools/android-tools/android-tools_5.1.1.r37.bb
```diff
@@ -92,7 +92,7 @@ do_compile() {
 
     # Setting both variables below causing our makefiles to not work with
     # implicit make rules
-    unset CFLAGS
+    CFLAGS="--file-prefix-map=${S}=/usr/src/debug/${PN}/${EXTENDPE}${PV}-${PR}"
```
kraj, line 95: This should be `-ffile-prefix-map`

kraj (changes requested):

meta-oe/recipes-devtools/android-tools/android-tools_5.1.1.r37.bb
chris2511: Changed.
Interestingly `--file-prefix-map` also works, but `-ffile-prefix-map` is the documented option.

kraj:
merged via 3baecf3983

PR #724, opened 2023-07-27
meta-python: add python3-telnetlib3 package

chase-qi:
telnetlib3 is a Telnet client and server library for python.


kraj:
It has been accepted into master via fc0e3cb91e

PR #725, opened 2023-07-28
image_types_sparse: Fix syntax error

chrisdimich:
When using the image type:

	IMAGE_FSTYPES += " wic.sparse"
	IMAGE_CLASSES += " image_types_sparse"

The following error arises:

	Syntax error: Bad function name

So need to remove function in favor of variable.

### DIFF
--- a/meta-oe/classes/image_types_sparse.bbclass
+++ b/meta-oe/classes/image_types_sparse.bbclass
@@ -8,9 +8,10 @@ inherit image_types
 SPARSE_BLOCK_SIZE ??= "4096"
 
 CONVERSIONTYPES += "sparse"
-CONVERSION_CMD:sparse() {
-    INPUT="${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}"
-    truncate --no-create --size=%${SPARSE_BLOCK_SIZE} "$INPUT"
-    img2simg -s "$INPUT" "$INPUT.sparse" ${SPARSE_BLOCK_SIZE}
-}
+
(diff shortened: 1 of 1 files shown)


meta-oe/classes/image_types_sparse.bbclass
```diff
@@ -8,9 +8,10 @@ inherit image_types
...
 
 CONVERSIONTYPES += "sparse"
-CONVERSION_CMD:sparse() {
-    INPUT="${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}"
-    truncate --no-create --size=%${SPARSE_BLOCK_SIZE} "$INPUT"
-    img2simg -s "$INPUT" "$INPUT.sparse" ${SPARSE_BLOCK_SIZE}
-}
+
+CONVERSION_CMD:sparse = " \
+    truncate --no-create --size=%${SPARSE_BLOCK_SIZE} "${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}"; \
```
shr-project, line 13: Does this even work in master (I see this PR is for mickledore, but if you're using this bbclass can you please check in master as well)?

I guess it needs to be updated for changes from:
https://git.openembedded.org/openembedded-core/commit/?id=26d97acc71379ab6702fa54a23b6542a3f51779c

first.

kraj (changes requested):
Please port it to master first and then a backport into `mickledore` would be in order.

chrisdimich:
Sure, I will port/test on master first.

PR #726, opened 2023-08-01
image_types_sparse: Fix syntax error

chrisdimich:
When using the image type:

	IMAGE_FSTYPES += " wic.sparse"
	IMAGE_CLASSES += " image_types_sparse"

The following error arises:

	Syntax error: Bad function name

So need to remove function in favor of variable.

Also remove IMAGE_NAME_SUFFIX as per:

https://git.openembedded.org/openembedded-core/commit/?id=26d97acc71379ab6702fa54a23b6542a3f51779c


kraj:
applied with dff205f5a3b15cbe1d63c5e351dae7a659042213

PR #738, opened 2023-09-07
microcom: add new recipe

ejoerns:
A minimalistic terminal program for communicating with devices over a serial connection.

It features connection via RS232 serial interfaces (including setting of transfer rates):

    microcom --speed=115200 --port=/dev/ttyS0

as well as in `telnetmode' as specified in rfc2217

    microcom --telnet=somehost:port

and a (Linux specific) can mode:

    microcom --can=interface:rx_id:tx_id


kraj:
applied to master with 2aa8abc917

Issue #744, opened 2023-10-05
please remove python3-botocore and python3-boto3


thomas-roos:
Hi,
we are maintaining meta-aws and those two recipes were recently 
https://github.com/openembedded/meta-openembedded/blob/master/meta-python/recipes-devtools/python/python3-botocore_1.31.57.bb
https://github.com/openembedded/meta-openembedded/blob/master/meta-python/recipes-devtools/python/python3-boto3_1.28.57.bb
added causing issues with https://bugzilla.yoctoproject.org/show_bug.cgi?id=15223
Discussed with Richard Purdie and he said people should just use ours - you might have not seen this.
We do ptests and daily upgrades for those recipes.
-> https://github.com/aws4embeddedlinux/meta-aws/tree/master/recipes-devtools/python
Thank you!
Cheers, Thomas

kraj:
yeah I am fine with it, Please send a pull to delete them with reasons as discussed.

kraj:
done

PR #745, opened 2023-10-06
python3-boto3, python3-botocore: remove recipes

thomas-roos:
python3-boto3 and python3-botocore are provided by meta-aws. They have ptests and upgraded as soon as there is an release. Having them here also causing issues with yocto-check-layer https://bugzilla.yoctoproject.org/show_bug.cgi?id=15223 As discussed on IRC with Richard Purdie, those recipes should just be provided by meta-aws. Agreed here https://github.com/openembedded/meta-openembedded/issues/744 by Khem.


kraj:
applied with bc98fb0765

PR #746, opened 2023-10-06
open62541: add Backport status and link to patch

JohannesKauffmann:


kraj:
applied with 0287a38fbc

PR #762, opened 2023-11-15
libebml: Enable shared libraries

Hains:
To fix build Gerbera, as:

  /usr/src/debug/libmatroska/1.7.1-r0/src/KaxBlock.cpp:538:(.text+0x4144):
  undefined reference to `libebml::SafeReadIOCallback::EndOfStreamX::EndOfStreamX(unsigned int)'


kraj:
applied with 237ce297fa

PR #765, opened 2023-11-21
open62541: update to v1.3.8

JohannesKauffmann:


kraj:
applied with 6850efd066

PR #768, opened 2023-12-01
glog: Disable 64bit atomics on armv{5,6}

williamspatrick:
The ARMv5 and ARMv6 architectures do not support 64-bit atomics, so
disable them.

Signed-off-by: Patrick Williams <patrick@stwcx.xyz>


kraj:
applied with d90c134f44

PR #773, opened 2024-01-08
abseil-cpp: force use of -std=c++14

thomas-roos:
Protobuf is compiled with -std=c++14, so abseil must be as well. Details can be found here: https://github.com/protocolbuffers/protobuf/issues/10768


shr-project:
It was switched to c++17 recently in:
https://git.openembedded.org/meta-openembedded/commit/?id=190f7b50ac3d7d30cf9d31167fb74f7146c3c8ff

if you want to revert this you should discuss on the ML.

PR #778, opened 2024-01-26
kernel-selftest: Add few more testcases

cshilwant:
This commit will build and install following testcases under /usr/kernel-selftest of filesystem,
    * ptp
    * timers


kraj:
applied to master with 02806ed4c1

PR #780, opened 2024-01-26
Update to uutils coreutils 0.24

Ecordonnier:


kraj:
applied to master with 1d125e1779 and 0bea9feffc

PR #791, opened 2024-02-16
python3-pybind11: Remove the Boost dependency

manvarpratik:
Boost is an optional dependency and it is only used in tests which are disabled by default in this recipe.

References:
https://github.com/pybind/pybind11/blob/master/tests/CMakeLists.txt#L325

Moreover, the Boost is an enormously large and complex suite of utility libraries that cost the filesystem memory footprint of Embedded devices.


kraj:
it should go to master branch first, I have cherry-picked to staging branch ( master-next ) into master since it is a simple cherry-pick, in future always check if the patch is needed on master branch first.

manvarpratik:
> it should go to master branch first, I have cherry-picked to staging branch ( master-next ) into master since it is a simple cherry-pick, in future always check if the patch is needed on master branch first.

Sure, I will keep this in mind. Thank you!

kraj:
applied to master with 086eeb45d4

PR #792, opened 2024-02-23
python3-gevent: adding missing dependency to python3-zopeevent

joergho:
... and python3-zopeinterface

As of [gevent/setup.py](https://github.com/gevent/gevent/blob/a67891991c95b5ffca78dc1f1a11f956ec3963ca/setup.py#L232), gevent depends on zope.event and zope.interfaces -> add those runtime dependencies to the recipe.


joergho:
Is it okay to backport this to kirkstone (latest LTS)? If yes, I'd be happy to create another PR for this.

kraj:
> Is it okay to backport this to kirkstone (latest LTS)? If yes, I'd be happy to create another PR for this.

yes it should be.

kraj:
applied to master with a5b340a387

PR #794, opened 2024-02-26
python3-pynacl: allow -native build

amuetzel:
This library is a dependency of other packages which we want to use in a native/sdk context. So I propose adding the necessary `BBCLASSEXTEND` to make it buildable here.


amuetzel:
If this kind of chanage is acceptable in principle, I would send separate PRs for three more recipes (python3-{bcrypt,paramiko,robotframework}). Or should I add them to this PR instead?

kraj:
applied to master with 067b6cd59b

PR #799, opened 2024-03-15
Revert back to stable versions of libqmi and libmbim

gotthardp:
The odd version numbers of libqmi and libmbim are unstable/development releases.
With the Scarthgap behind the door I'd really suggest to use the stable releases only as we used to.


kraj:
applied with b7d577d8d3 and 538b03a3ff

PR #818, opened 2024-04-30
meta-oe(collectd): add package config option to build with mqtt plugin

reubenmiller:
Add package config option for the `collectd` recipe to include the MQTT plugin in the collectd package.

This enables collectd to publish its metrics to MQTT broker where the metrics can be subscribed to by other components.


reubenmiller:
Sorry I didn't see that patches should be sent via the mailing list.

PR #819, opened 2024-04-30
networkd-dispatcher: Add dependency on python3-json

maffan:
The `networkd-dispatcher` script imports json. Add it as an RDEPENDS.


kraj:
applied with 4fc97fe602

PR #824, opened 2024-05-05
python3-twisted: Fix broken package split V2

Hains:
Without FILES:${PN} only main, dbg, dev, doc and test packages are being build. python3-twisted-core, python3-twisted-web, python3-twisted-conch, etc are missing.

Remove newsframents directory as it only contains a .gitignore file for development, thereby silence warning:

WARNING: python3-twisted-24.3.0-r0 do_package: QA Issue: python3-twisted: Files/directories were installed but not shipped in any package:
  /usr/lib/python3.12/site-packages/twisted/newsfragments
  /usr/lib/python3.12/site-packages/twisted/newsfragments/.gitignore

### DIFF
--- a/meta-python/recipes-devtools/python/python3-twisted_24.3.0.bb
+++ b/meta-python/recipes-devtools/python/python3-twisted_24.3.0.bb
@@ -11,8 +11,9 @@ SRC_URI[sha256sum] = "6b38b6ece7296b5e122c9eb17da2eeab3d98a198f50ca9efd00fb03e5b
 inherit pypi python_hatchling
 
 do_install:append() {
-    # remove some useless files before packaging
+    # remove some useless files and newsfragments before packaging
     find ${D} \( -name "*.bat" -o -name "*.c" -o -name "*.h" \) -exec rm -f {} \;
+    rm -rf ${D}${PYTHON_SITEPACKAGES_DIR}/twisted/newsfragments
 }
 
 PACKAGES += "\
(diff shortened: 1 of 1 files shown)


gudnimg:
I was curious and looked a bit today into _why_ this is happening where `FILES:${PN}` is explicitly required, while in some recipes its not required at all.

**TLDR:** This oneliner will fix the issue `PACKAGES += "\` changes to `PACKAGES =+ "\` (I can submit a patch for this if you prefer)

----

I found that `setuptools3-base.bbclass` defines: `FILES:${PN} += "${PYTHON_SITEPACKAGES_DIR}"`

And with your patch you're basically overwriting `FILES:${PN}` with `=` which doesn't solve the underlying problem because now any file which is not picked up by the split packages like python3-twisted-core will now raise a warning like you see with `newsfragment`.

On my end, it works simply to change the PACKAGES definition by using `=+` (prepend) instead of `+=` (append) which seems to be quite common in other recipes. This way, the packages ${PN}-conch,${PN}- web, ${PN}-core etc.  will come before `${PN}`.

This error will also disappear since it will now be included in `FILES:${PN}`:
```
WARNING: python3-twisted-24.3.0-r0 do_package: QA Issue: python3-twisted: Files/directories were installed but not shipped in any package:
/usr/lib/python3.12/site-packages/twisted/newsfragments
/usr/lib/python3.12/site-packages/twisted/newsfragments/.gitignore
```

----

Just to show it works on my end :) I'm building https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-core/images/core-image-minimal.bb and just adding `IMAGE_INSTALL:append = "python3-twisted"`

image

Hains:
Tested and changing packaging from append to prepend does solve the issue, without warnings and errors..
Please submit patch.

gudnimg:
Patch sent: https://lists.openembedded.org/g/openembedded-devel/topic/meta_python_patch/105930078

PR #830, opened 2024-06-05
cukinia: upgrade 0.6.2 -> 0.7.0

deribaucourt:


deribaucourt:
Changes:
        cukinia: Add user is member of groups function
        cukinia: correct _ver2int output format
        cukinia: filter junitxml output to only include valid xml 1.0 chars
        cukinia: Add network interface functions
        cukinia: correct log_suite file name
        cukinia: differentiate cukinia_gpio_* tests
        cukinia_i2c: check driver
        add when/unless
        cukinia: fix bashism in result comparison
        Change license to Apache-2.0 only

    GPL-3.0-only license has been removed intentionally (see changelog).

No additional RRECOMMENDS for the new features from looking at the code (only depends on ip).

kraj:
applied with bfc372f8c49e735782287197c2ec78bb5d233342

PR #831, opened 2024-06-05
udpcast: add recipe

deribaucourt:
udpcast is a file transfer tool that can send data simultaneously to many destinations on a LAN.

It is an alternative to uftp which is already in meta-networking.

On a lossy network, udpcast provided 10x faster transfer rates with error correction enabled while using a half as much CPU thanks to a simpler algorithm.

### DIFF
--- a/meta-networking/recipes-core/packagegroups/packagegroup-meta-networking.bb
+++ b/meta-networking/recipes-core/packagegroups/packagegroup-meta-networking.bb
@@ -236,6 +236,7 @@ RDEPENDS:packagegroup-meta-networking-support = "\
     rdma-core \
     tcpreplay \
     tinyproxy \
+    udpcast \
     uftp \
     unbound \
     vnstat \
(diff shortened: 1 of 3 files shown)


meta-networking/recipes-support/udpcast/udpcast/0001-don-t-depend-on-pod2man-to-build.patch
```diff
@@ -0,0 +1,40 @@
+From 77f6e709a350770fada77e9f543e4e927ce44071 Mon Sep 17 00:00:00 2001
+From: Enguerrand de Ribaucourt <enguerrand.de-ribaucourt@savoirfairelinux.com>
+Date: Mon, 20 May 2024 16:36:13 +0200
+Subject: [PATCH] don't depend on pod2man to build
+
+Upstream-Status: Inappropriate
```
kraj, line 6: it will be good to specify reason why it is Inappropriate,

kraj (changes requested):

kraj:
it fails to build on musl so please address that too - https://errors.yoctoproject.org/Errors/Details/785867/

kraj:
added with 8b01302835

PR #849, opened 2024-07-23
nginx: add PACKAGECONFIG knobs for fastcgi, scgi and uwsgi

maxinbjohn:
fastcgi, scgi and uwsgi are enabled by default in nginx. Provide an option to disable these features (that reduces binary size by 8%).


kraj:
fixed with a944926d1931bc931bacd90f19824a427408bb40

PR #850, opened 2024-07-25
Improve polkit recipe

mferland:
This merge request:

1. updates the SRC_URI since the project was moved to GH
2. Adds the elogind and libs-only PACKAGECONFIG options
3. Makes backend PACKAGECONFIG options mutually exclusive


kraj:
applied the series at 30e64975d302d69c54e56dee964e73da633e2aa9

PR #851, opened 2024-07-29
android-tools-adbd.service: Change /var to /etc in ConditionPathExists

quic-raghuvar:
If android-tools-adbd.service service needs to be up upon boot, then the path assigned to ConditionPathExists must be present at boot time. This means that the path set to ConditionPathExists must be created at build time itself. In this case it is var/usb-debugging-enabled. However, /etc is a better place to keep files and directories that are created at build time rather than /var. /var is expected to house files that are created at run time.

Hence, change ConditionPathExists=/var/usb-debugging-enabled to ConditionPathExists=/etc/usb-debugging-enabled

This helps enablement of android-tools-adbd.service from build by creating /etc/usb-debugging-enabled at build time.

### DIFF
--- a/meta-oe/recipes-devtools/android-tools/android-tools/android-tools-adbd.service
+++ b/meta-oe/recipes-devtools/android-tools/android-tools/android-tools-adbd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Android Debug Bridge
-ConditionPathExists=/var/usb-debugging-enabled
+ConditionPathExists=/etc/usb-debugging-enabled
 Before=android-system.service
 
 [Service]


kraj:
applied to master with eb31674b1c050ebccf0c85f4accc15cc7ea4b58a

quic-raghuvar:
@kraj Can this patch be pulled to krikstone branch as well?

kraj:
Send a pr against Kirkstone or better on mailing list

quic-raghuvar:
Done. Created a PR against Kirkstone. [https://github.com/openembedded/meta-openembedded/pull/857](url)

lumag:
@kraj you have merged the patches (this one and #857), however they don't have DCO (Signed-off-by), they have a lame name in the Author field ("quic-raghuvar") instead of the full name, the commit messages is a text dump which is not wrapped on the useful text width.

Please consider putting more attention to the GitHub PRs, they can come from undereducated developers and can have mistakes which are easy to miss thanks to the GH interface.

kraj:
@lumag thanks for noting it, I think it slipped in due to oversight. I think because this pull pushed the change after the effect. 
@quic-raghuvar please follow

https://docs.yoctoproject.org/dev/contributor-guide/submit-changes.html

to prepare patches with good commit msg and patch headers in future.

lumag:
@kraj maybe we should add GitHub actions to do some checks (at least format, DCO, etc). Maybe we can use TuxSuite / TuxBake to perform build-testing. This should help reviewers.

PR #852, opened 2024-08-01
libcamera: Use multiple of sizeof as malloc size

hubertwisniewski0:
Mallocs being the subject of this patch replace VLAs (patch `0001-rpi-Use-alloca-instead-of-variable-length-arrays.patch`), so they should allocate memory in blocks whose sizes are multiples of the respective types' sizes.


kraj (commented):
Please submit this patch to upstream libcamera as well and add `Upstream-Status` tag to the patch based upon the feedback.

hubertwisniewski0:
This patch fixes the problem introduced by patch `0001-rpi-Use-alloca-instead-of-variable-length-arrays.patch`, which has been rejected by upstream [[PATCH] rpi: Use alloca instead of variable length arrays](https://lists.libcamera.org/pipermail/libcamera-devel/2024-February/040536.html) but is still here anyway (with Upstream-Status: Submitted). So, either this patch should be added or the `0001` patch should be removed (or perhaps they should be merged and submitted to upstream?).

kraj:
> This patch fixes the problem introduced by patch `0001-rpi-Use-alloca-instead-of-variable-length-arrays.patch`, which has been rejected by upstream [[PATCH] rpi: Use alloca instead of variable length arrays](https://lists.libcamera.org/pipermail/libcamera-devel/2024-February/040536.html) but is still here anyway (with Upstream-Status: Submitted). So, either this patch should be added or the `0001` patch should be removed (or perhaps they should be merged and submitted to upstream?).

in that case just do this changes in original patch and mark the patch as "rejected" we need to restart submission with v2

PR #854, opened 2024-08-05
Include UPSTREAM_CHECK_COMMITS to fix UNKNOWN_BROKEN status when running devtool check-upgrade-status

altru-smile:
When performing devtool check-upgrade-status, UNKNOWN_BROKEN status appears.
On the upstream source repository, releases are not identified by tags. So,
UPSTREAM_CHECK_COMMITS is set to 1, to find the latest upstream update.

cc: @ycongal-smile


kraj:
applied at 9c8045fad081ff5e25a74e4b2d23bcb4dd56259a

PR #857, opened 2024-08-07
android-tools-adbd.service: Change /var to /etc in ConditionPathExists

quic-raghuvar:
If android-tools-adbd.service service needs to be up upon boot, then the path assigned to ConditionPathExists must be present at boot time. This means that the path set to ConditionPathExists must be created at build time itself. /etc is a better place to keep files and directories that are created at build time rather than /var. /var is expected to house files that are created at run time.

Hence, change ConditionPathExists=/var/usb-debugging-enabled to ConditionPathExists=/etc/usb-debugging-enabled


kraj:
applied with 8106cfe769aa3f0ebdff7a1caed018f9ab49a090

PR #859, opened 2024-08-12
polkit: update to 125

bluca:


meta-oe/recipes-extended/polkit/polkit-group-rule.inc
```diff
@@ -5,9 +5,3 @@ inherit features_check
 REQUIRED_DISTRO_FEATURES = "polkit"
 
 inherit useradd
```
petermarko, line 7: Below you remove the only useradd parameters, so there is no user created/modified anymore.

Therefore this inherit should be removed too, together with
recipes-extended/polkit/polkit-group-rule-datetime.bb:USERADD_PACKAGES = "${PN}"
recipes-extended/polkit/polkit-group-rule-network.bb:USERADD_PACKAGES = "${PN}"
recipes-extended/polkit/polkit-group-rule-udisks2.bb:USERADD_PACKAGES = "${PN}"

bluca: Those recipes are adding new groups, so I think it's still needed? eg: `GROUPADD_PARAM:${PN} = "--system network"`

petermarko: OK, sorry for the noise.

shr-project:
What is supposed to package ${datadir}/polkit-1? unlike ${sysconfdir} it's not in default FILES variables causing:

```
ERROR: QA Issue: polkit-group-rule-network: Files/directories were installed but not shipped in any package:
  /usr
  /usr/share
  /usr/share/polkit-1
  /usr/share/polkit-1/rules.d
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
polkit-group-rule-network: 4 installed and not shipped files. [installed-vs-shipped]
```

bluca:
I thought that was picked up by the polkit package, maybe I misunderstood how it's used - can you try adding something like `FILES:${PN} += "${datadir}/polkit-1/rules.d" to meta-oe/recipes-extended/polkit/polkit-group-rule.inc and see if that makes it happy?

shr-project:
there is no polkit package created by these recipes.

FILES:${PN} += "${datadir}/polkit-1/rules.d"
in the .inc files should work, I haven't tested it as I don't use any of these recipes, it was just new failure shown in my world builds, didn't you test this change before sending it?

bluca:
> there is no polkit package created by these recipes.

Yeah, it's in the other recipe in the same directory, hence the confusion

> didn't you test this change before sending it?

As part of a full OS build for the new polkit version, but didn't think of checking whether these were running too, I guess they are not pulled in by that config

PR #861, opened 2024-08-13
Upgrade dbus-cxx to 2.5.2

rm5248:
In addition, allow for other configuration options for dbus-cxx and split into different packages.

SDK can now also be built with dbus-cxx tools, generally used to generate C++ interface code from DBus introspection files.  This required adding cppgenerate in a new recipe for dependency purposes.


kraj:
applied to master branch at 5df4beeb75

PR #865, opened 2024-08-29
uutils-coreutils: disable buildpaths error

Ecordonnier:


shr-project:
Don't just disable it, if it cannot be fixed, then lower it from ERROR_QA to WARN_QA (so that it's still shown until someone fixes it properly).

Ecordonnier:
@shr-project I force-pushed the PR with a new version implementing your feedback.

kraj:
I would encourage you to root cause it, I know its rust and perhaps somewhere rust compiler not respecting debug-prefix mapping will be at the heart of the problem. I am trying to clean meta-openembedded for reproducibility and buildpaths is important to get rid of for that.

shr-project:
I didn't check uutils itself, but these:
https://github.com/rust-lang/rust/labels/A-reproducibility
https://github.com/rust-lang/rust/issues/111540
might be useful here as well.

Ecordonnier:
I would also like to root-cause when I have some time. I thought in the meanwhile this would be useful to fix the immediate build failure. If you prefer to wait for a proper fix, it's fine to not merge this PR, however at the moment I can't commit to a timeline for a proper fix.

Ecordonnier:
Merged as https://github.com/openembedded/meta-openembedded/commit/deca9342abf136db4bbac921ca28a98c0cf032cd

Ecordonnier:
Upstream issue: https://github.com/uutils/coreutils/issues/6833

PR #869, opened 2024-09-10
postgresql: use PACKAGECONFIG flag for readline dependency

slpn-lampe:
This patch moves the dependency on readline from the package's `DEPENDS` variable into a `PACKAGECONFIG[readline]`option and adds `readline`to the default `PACKAGECONFIG`.

This allows a global override `PACKAGECONFIG:remove = "readline"`to remove readline support and runtime dependency from the psql client.


kraj:
applied to master @ ba274e7310

PR #870, opened 2024-09-12
genimage: add new recipe

ejoerns:
This adds support for the files system and disk image creation tool [genimage](https://github.com/pengutronix/genimage) to meta-oe.

This can be an alternative for use cases where the oe-core tooling has limitations.
E.g. when building and re-using intermediate file system images or in the context of more complex disk image creations.

Since *genimage* uses *libconfuse*, which was already part of meta-networking, this also moves libconfuse to meta-oe and prepares it for use from genimage.


meta-oe/recipes-support/genimage/genimage_18.bb
```diff
@@ -0,0 +1,3 @@
+require genimage.inc
```
kraj, line 1: .inc files are only needed when there are more than 1 recipes needing them so please merge it into .bb

kraj (changes requested):
have you tried proposing it to OE-core ?

ejoerns:
> have you tried proposing it to OE-core ?

No, not yet.
In `oe-core` there is `wic` which is a bit comparable to what `genimage` provides. However, `wic` itself and the way it is used in oe-core have their limitations. Nevertheless, my impression is that oe-core generally seems to be very careful not to accept recipes that could be similar to existing solutions. Thus I thought that `meta-oe` would better fit for providing `genimage`.

kraj:
applied with c3a1158917

PR #877, opened 2024-10-02
netdata: Support more plugins

deribaucourt:
This PR introduces a few commits which improve our support for the Netdata recipe.

I had to disable a few QA checks when building the go plugin. Let me know if these workarounds are not acceptable, and we should instead find solutions by fixing the code. I tested some go plugins at runtime and did not encounter any error despite the warnings being disabled.


kraj:
looks ok from quick looks,  I have staged it into `master-next` for more CI

kraj:
applied to master @ d5a24aa9ba

PR #882, opened 2024-10-09
fitimage: add support to build arbitrary FIT images

ejoerns:
The FIT image support in OE is quite limited:
 1) No support to build an arbitrary number of FIT images since the FIT
    image generation is tightly coupled to the kernel image.
 2) A lot of U_BOOT-specific variables which may not be necessary for
    other bootloaders.
 3) No usage of the meta-oe signing.bbclass for signed FIT images.

This alternative class is added to solve the above-mentioned problems:
 1) The class can be inherited by an arbitrary number of
    `<fit-image-name>.bb` recipes to generate FIT images
 2) No U_BOOT-specific variables are used
 3) `<fit-image-name>.bb` recipes can prepend the `do_fitimage()` to
    provide the key using the [signing.bbclass](https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/classes/signing.bbclass) e.g.:

        do_fitimage:prepend() {
            signing_prepare
            signing_use_role "${FITIMAGE_SIGNING_KEY_ROLE}"
        }

    Then enable and configure signing as follows:

        FITIMAGE_SIGN = "1"
        FITIMAGE_MKIMAGE_EXTRA_ARGS = "--engine pkcs11"
        FITIMAGE_SIGN_KEYDIR = "${PKCS11_URI}

This class is inspired by the meta-phytec fitimage.bbclass [1].

[1] https://git.phytec.de/meta-phytec/tree/classes/fitimage.bbclass

### DIFF
--- a/meta-oe/classes/fitimage.bbclass
+++ b/meta-oe/classes/fitimage.bbclass
@@ -1,0 +1,530 @@
+# SPDX-License-Identifier: MIT
+#
+# Copyright PHYTEC Messtechnik GmbH
+# Copyright (C) 2024 Pengutronix, <yocto@pengutronix.de>
+#
+# Class for creating (signed) FIT images
+# Description:
+#
+# You have to define the 'images' to put in the FIT image in your recipe file
+# following this example:
+#
+#    FITIMAGE_IMAGES ?= "kernel fdt fdto setup ramdisk bootscript"
+#
+#    FITIMAGE_IMAGE_kernel ?= "virtual/kernel"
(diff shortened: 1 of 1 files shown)


ejoerns:
Added `FITIMAGE_CONFIG_PREFIX ?= "conf-"`as prefix for configuration sections to not collide with possible dtb sections and better align with the oe-core pendant to ease migration if desired.

kraj:
applied with b20be52a4e

PR #883, opened 2024-10-11
Move efitools & sbsigntool recipes from meta-secure-core

jetm:
`efitools` is coming from meta-secure-code. There are many cases where meta-secure-code is added just to use `efitools`, like https://github.com/Wind-River/meta-secure-core/pull/76

`efitools` has a dependency on `sbsigntool`.

### DIFF
--- a/meta-oe/classes/user-key-store.bbclass
+++ b/meta-oe/classes/user-key-store.bbclass
@@ -1,0 +1,546 @@
+DEPENDS:append:class-target = " \
+    ${@bb.utils.contains("DISTRO_FEATURES", "efi-secure-boot", "sbsigntool-native", "", d)} \
+    ${@bb.utils.contains("DISTRO_FEATURES", "efi-secure-boot", "libsign-native", "", d)} \
+    openssl-native \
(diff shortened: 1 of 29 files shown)


jetm:
Tagging @yizhao1 to keep it in the loop

quaresmajose (approved):
LGTM

kraj:
I would like to see some activity on the patches, there are many in these recipes and none of them are upstream or on its way to upstream, its quite a bit to carry if we apply it here, since it has a chance to fall back on maintainers back. Please submit these patches upstream and get some feedback

jetm:
Hi @kraj,

> I would like to see some activity on the patches, there are many in these recipes and none of them are upstream or on its way to upstream, its quite a bit to carry if we apply it here, since it has a chance to fall back on maintainers back. Please submit these patches upstream and get some feedback

The upstream activity can be checked out at https://github.com/Wind-River/meta-secure-core/commits/master/meta-efi-secure-boot/recipes-bsp/efitools. Both recipes have been widely tested. This is just a copy of them from meta-secure-core.

I don't understand where to submit the patches. Which upstream?

jetm:
Sending meta-secure-core efitool patches to efitool upstream to get feedback from efitool maintainer

PR #891, opened 2024-10-24
netdata: adjust xenstat.plugin and app.plugin permissions, and add virtual/docker RDEPENDS

TanguyRaufflet:
3 modifications are done in this PR, that improves the netdata recipe:
- add setuid for the xenstat plugin and change the permission access
- change the permission access of the file app.plugin to fit with xenstat.plugin, which is more secure
- add virtual/docker RDEPENDS in the docker packageconfig


kraj:
applied to master with 8a34c94412

PR #901, opened 2024-11-17
uutils-coreutils: upgrade 0.0.27 -> 0.0.28

Ecordonnier:
See https://github.com/uutils/coreutils/releases/tag/0.0.28


kraj:
it was also proposed on ml
applied with 2866036261

PR #903, opened 2024-11-20
bmap-writer: add initial recipe

embetrix:
this is an utility designed to efficiently write disk images
to storage devices using block mapping (BMAP).
It serves as a lightweight alternative to the Yocto BMAP tool.
it is C++ based does not require Python and focuses solely on writing an image.


kraj:
merged with 7cf3b1535c

PR #909, opened 2025-01-02
ctre: add new recipe

xbrtszstk:
Hello,
I would like to add a new recipe to the meta-oe layer.
This recipe is another component required by the reflect-cpp project, which I want to add to this layer.


kraj:
applied with 21ce886005

PR #913, opened 2025-01-02
bmap-writer: update to latest git version

embetrix:
* switch to libarchive to handle a larger decompression schemes

* implement own sha256 hashing and drop openssl dependency

* compute maximum buffer size before writing each range

* bmap file optional: if not provided, it will be searched in the same path as the input

* print time/speed of the writing operation


kraj:
applied with 8bdd4740b2

PR #914, opened 2025-01-02
botan: update 3.5.0 -> 3.6.1

embetrix:
* update to latest 3.6.1 Version
* add packaging for botan binary and botan-test tool

### DIFF
--- a/meta-oe/recipes-crypto/botan/botan_3.6.1.bb
+++ b/meta-oe/recipes-crypto/botan/botan_3.6.1.bb
@@ -5,7 +5,7 @@ LIC_FILES_CHKSUM = "file://license.txt;md5=573e74513ae3057b04757df65b537de0"
 SECTION = "libs"
 
 SRC_URI = "https://botan.randombit.net/releases/Botan-${PV}.tar.xz"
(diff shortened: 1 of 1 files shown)


kraj (changes requested):
Patches need to be forward ported and Upstream-Status added see
https://valkyrie.yoctoproject.org/#/builders/81/builds/245/steps/15/logs/errors

embetrix:
@kraj the mainline project does not consider the installation of the `botan-test` binary and data but meant to run only to run on the host after the build. 
what do you suggest ? can we have skip the QA for this patch ?

kraj:
> @kraj the mainline project does not consider the installation of the `botan-test` binary and data but meant to run only to run on the host after the build. what do you suggest ? can we have skip the QA for this patch ?

If we are not using the test app then we can try to not build it and side step the issue

embetrix:
@kraj : I dropped the patch, the test data can be also specified from `botan-test` command line argument:

`botan-test --data-dir=/usr/share/botan/tests/data`

kraj:
applied with 338e2ff78b

PR #915, opened 2025-01-05
polkit: Install rules in subdir

MagneFire:
https://github.com/openembedded/meta-openembedded/commit/d5e90541f8e35916abc930b2da6de037b23d51a1 moved the rules to /usr/share/ instead of /etc/. The commit also removed the install:prepend() step.
This results in the rules being installed as file /usr/share/polkit-1/rules.d instead of in that folder.

This commit adds back the install prepend step such that the rules are installed in said folder.


kraj:
applied with d89fc818b7

Issue #916, opened 2025-01-10
python3-posix-ipc builds without posix memory queue support


AleksandrVin:
There is an issue during the Python3 package build. A prober.py script is meant to be run during the build process [posix-ipc](https://github.com/osvenskan/posix_ipc/blob/rel1.1.1/prober.py).  package: python3-posix-ipc

Using the dev shell, it runs normally, though using receipt from open embedded builds without memory queue support, so the Python scripts fail with an appropriate error when run in qemu.  

```python3
  File "/usr/lib/python3.12/site-packages/dd/service/receiver.py", line 18, in __init__
    self.message_queue = MessageQueueRepository(queue_name=queue_name)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/dd/repository/file.py", line 152, in __init__
    self.queue = posix_ipc.MessageQueue(queue_name)
                 ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'posix_ipc' has no attribute 'MessageQueue'
```

PR #917, opened 2025-01-11
fix missing posix message queue support in python3-posix-ipc package

AleksandrVin:
[issue](https://github.com/openembedded/meta-openembedded/issues/916)

There was a problem with this package, it compiled normally, but without message queue support, even though it was enabled due to the latest kernel usage. 

Without DEPENDS += "glibc" [probber script](https://github.com/osvenskan/posix_ipc/blob/rel1.1.1/prober.py) tends to provide wrong results, disabling message queue support. 

This hack fixes it. I'm new to open embedded, so please tell me if this approach is not the optimal one or not correct.

### DIFF
--- a/meta-python/recipes-devtools/python/python3-posix-ipc_1.1.1.bb
+++ b/meta-python/recipes-devtools/python/python3-posix-ipc_1.1.1.bb
@@ -10,3 +10,7 @@ UPSTREAM_CHECK_PYPI_PACKAGE = "${PYPI_PACKAGE}"
 SRC_URI[sha256sum] = "e2456ba0cfb2ee5ba14121450e8d825b3c4a1461fca0761220aab66d4111cbb7"
 
 inherit setuptools3 pypi
+
+# Ensure posix message queue support in python package for prober.py script
+DEPENDS += "glibc"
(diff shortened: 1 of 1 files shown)


meta-python/recipes-devtools/python/python3-posix-ipc_1.1.1.bb
```diff
@@ -10,3 +10,7 @@ UPSTREAM_CHECK_PYPI_PACKAGE = "${PYPI_PACKAGE}"
 SRC_URI[sha256sum] = "e2456ba0cfb2ee5ba14121450e8d825b3c4a1461fca0761220aab66d4111cbb7"
 
 inherit setuptools3 pypi
+
+# Ensure posix message queue support in python package for prober.py script
+DEPENDS += "glibc"
```
kraj, line 15: glibc is default dependency added to every recipe if the distro you are building is using `TCLIBC = "glibc"` so I wonder if you are using TCLIBC = "musl" ?

AleksandrVin: hmm, yes. it's not a solution

PR #918, opened 2025-01-13
pistache : add initial recipe

embetrix:
Add `pistache` library a high-performance REST toolkit written in C++ :
https://github.com/pistacheio/pistache

### DIFF
--- a/meta-oe/recipes-devtools/pistache/pistache_0.4.26.bb
+++ b/meta-oe/recipes-devtools/pistache/pistache_0.4.26.bb
@@ -1,0 +1,44 @@
+SUMMARY = "A high-performance REST toolkit written in C++"
+HOMEPAGE = "pistacheio.github.io/pistache"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=fa818a259cbed7ce8bc2a22d35a464fc"
+
+SRC_URI = "git://github.com/pistacheio/pistache.git;branch=master;protocol=https"
+# SRCREV tagged 0.4.26
+SRCREV = "ddffda861aa49012dcda28f1362d0339e718cd52"
+
+S = "${WORKDIR}/git"
+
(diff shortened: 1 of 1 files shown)


kraj:
fails in do_install
https://errors.yoctoproject.org/Errors/Details/840103/

embetrix:
@kraj : problem fixed

kraj:
fails still - https://errors.yoctoproject.org/Errors/Details/840151/

embetrix:
@kraj it looks like this package is incompatible with musl :-(

embetrix:
@kraj : added `COMPATIBLE_HOST:libc-musl = "null`

OldManYellsAtCloud:
1. Those dependencies shouldn't be mandatory if they are not needed, e.g. brotli shouldn't be a dependency if brotli support is not included. The only mandatory dependency seems to be rapidjson, the rest are PACKAGECONFIG-dependent.
2. The install step fails, if "examples" is not part of PACKAGECONFIG
3. The "test" PACKAGECONFIG has a build-time dependency on curl and gtest, however these are not included, and the build fails, when it is enabled.

embetrix:
@OldManYellsAtCloud thanks for the input, I included your suggestion to https://github.com/openembedded/meta-openembedded/pull/926

@kraj : it seems that this MR is already merged ! I didn't get a notification...

OldManYellsAtCloud:
Regarding the dependencies: the next parameter in the PACKAGECONFIG value is the compile-time dependency that's added automatically, in case the config is enabled, e.g.
```
PACKAGECONFIG[openssl]  = "-DPISTACHE_USE_SSL=true, -DPISTACHE_USE_SSL=false, openssl"
```
This adds openssl to `DEPENDS` automatically, if openssl is enabled in PACKAGECONFIG

embetrix:
@OldManYellsAtCloud good hint, I change it again

PR #920, opened 2025-01-15
polkit: fix permissions of /usr/share/polkit-1/rules.d/ again

bluca:
Commit d89fc818b716d099f83a5a9a973be428e2b66806 changed the permissions back to 700, which is wrong for /usr/share, these files are intended to be world readable. Change it back.

Fixes: d89fc818b716 ("polkit: Install rules in subdir")


kraj:
applied to master with d37bf862241e0c6a6f3ec07c13b8898ac338629e

PR #923, opened 2025-01-16
reflect-cpp: add new recipe

xbrtszstk:


kraj:
fails when using llvm libc++ instead of gcc libstdc++

RISCV64

https://errors.yoctoproject.org/Errors/Details/840261/

ARM

https://errors.yoctoproject.org/Errors/Details/840263/

liuzicheng1987:
@xbrtszstk , we have automated build pipelines for LLVM, both on Linux as macOS, using several different versions of the LLVM compiler:

https://github.com/getml/reflect-cpp/actions/runs/12634925477

https://github.com/getml/reflect-cpp/actions/runs/12634925462

Needless to say, it compiles. 

So it must be something that is somehow specific to what you are doing. What version of the compiler and what flags are you using there?

kraj:
> liuzicheng1987



> @xbrtszstk , we have automated build pipelines for LLVM, both on Linux as macOS, using several different versions of the LLVM compiler:
> 
> https://github.com/getml/reflect-cpp/actions/runs/12634925477
> 
> https://github.com/getml/reflect-cpp/actions/runs/12634925462
> 
> Needless to say, it compiles.
> 
> So it must be something that is somehow specific to what you are doing. What version of the compiler and what flags are you using there?

We are using latest clang-19 and libc++ is from same, I think you will see similar issue with FreeBSD if its using clang 19. I am not sure about MacOS version of LLVM/Clang, I am reasonably sure that its not based on clang19 yet perhaps still using older releases.

liuzicheng1987:
All right...I'll try to add a build pipeline for Clang 19 and see what happens.

PR #933, opened 2025-02-08
[libspdm/spdm-utils] exclude Mipsel from build.

fairbird:


kraj:
merged at f0f690a64d into master

PR #934, opened 2025-02-12
libspdm: add support for ppc64le

geissonator:
Recently picked this up over in the OpenBMC project and noticed our ppc64le based server could no longer build the code.

Verified this change allows everything to build fine on ppc64le servers now.


kraj:
thanks and merged with f6352cd450 into master branch

PR #941, opened 2025-03-04
fsverity-utils: fix SRC_URI

Ecordonnier:
This URL does not exist any more, and do_fetch works only because a mirrored file is available at
http://downloads.yoctoproject.org/mirror/sources/git2_git.kernel.org.pub.scm.linux.kernel.git.ebiggers.fsverity-utils.git.tar.gz


kraj:
Fixed with d54e2671e6 on master

PR #944, opened 2025-03-07
python3-h5py: switch to PEP-517 build backend; don't enforce hdf5 lib version

okaestne:
Upstream moved to PEP-517 builds, so use the appropriate bbclass.

Also don't enforce hdf5 lib version, as it will be automatically detected.
This fixes the following runtime warning:

    >>> import h5py
    /usr/lib/python3.13/site-packages/h5py/__init__.py:36: UserWarning: h5py is running against HDF5 1.14.4 when it was built against 1.14.0, this may cause problems
    _warn(("h5py is running against HDF5 {0} when it was built against {1}, "


kraj:
applied to master at 8aa895e877

PR #945, opened 2025-03-07
uutils-coreutils: disable stdbuf compilation

Ecordonnier:
There is a bug in uutils-coreutils causing stdbuf to include the HOST version of libstdbuf.so, even when cross-compiled. As a side-effect, the uutils-coreutils binary was containing the buildpath, because RPATH was set on libstdbuf.so. Thus the buildpath error can be re-enabled.

This is tracked upstream by https://github.com/uutils/coreutils/issues/6591


kraj:
applied to master with 1d43511321

Issue #948, opened 2025-03-12
gitpkgv.bbclass: bitbake's fast initial shallow fetch


WXbet:
This bitbake commit https://github.com/openembedded/bitbake/commit/457288b2fda86fd00cdcaefac616129b0029e1f9 seems to break `gitpkgv.bbclass` here: https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/classes/gitpkgv.bbclass#L83

If your default is BB_GIT_SHALLOW = "1", the bare clone repo will not be created in the "sources/git2" folder (as before) and "gitpkgv.bbclass" returns **None**.

Possible solution is, to force BB_GIT_SHALLOW = "0" in that class:

![Image](https://github.com/user-attachments/assets/091a473a-0f83-4259-8646-120f171c0215)

PR #951, opened 2025-03-22
gitpkgv.bbclass: Fix broken git revison

Hains:
The revision dictionary was built with keys from names list in Bitbake(setup_revisions()). See:
https://git.openembedded.org/bitbake/commit/?id=2515fbd10824005fa7f34e87706000c079920366

And used to build names list in gitpkgv.bbclass in the old situation, see: https://git.openembedded.org/meta-openembedded/commit/?id=2920d4909236106e1a36d56b3b20762a308ba3d4

Use name variable to build names list instead of revision variable. Use append() now name variable is a string.

Old ipk file name:
enigma2_3.13+git3_1_3_0_9_c_7_0_a_4_f_a_6_9_d_6_7_0_1_7_b_b_9_f_7_f_4_d_d_9_d_6_1_0_a_8_c_3_d_20+31309c70a4-r0_dm920.ipk

New ipk file name:
enigma2_3.13+git21834+31309c70+31309c70a4-r0_dm920.ipk

### DIFF
--- a/meta-oe/classes/gitpkgv.bbclass
+++ b/meta-oe/classes/gitpkgv.bbclass
@@ -70,7 +70,7 @@ def get_git_pkgv(d, use_tags):
         names = []
         for url in ud.values():
             if url.type == 'git' or url.type == 'gitsm':
-                names.extend(url.revision)
+                names.append(url.name)
         if len(names) > 0:
             format = '_'.join(names)
         else:


WXbet:
... and we still need this https://github.com/openembedded/meta-openembedded/issues/948 to get it working if shallow cloning is enabled.

kraj:
> ... and we still need this #948 to get it working if shallow cloning is enabled.

Can you reply to Richard's patch on mailng list for meta-openembedded where he fixed it to work with new bitbake rework?

PR #953, opened 2025-03-28
networkmanager: only enable the vala bindings if GObject introspection is enabled

afenkart:
Fixes build errors in network-manager with GI_DATA_ENABLED = "False"


afenkart:
see here: https://github.com/openembedded/meta-openembedded/pull/954

PR #956, opened 2025-04-11
source-highlight: add recipe

Ecordonnier:
source-highlight version 3.1.8 was removed from openembedded-core in https://web.git.yoctoproject.org/poky/commit/?id=6560db9ddfd39a2715ede96430228306b00f53dd in June 2019 because the was no user left, however there is now a gdb PACKAGECONFIG option which uses it. Since the last release was in 2019, let's use the latest git version rather than the latest formal release.

### DIFF
--- a/meta-oe/recipes-support/source-highlight/source-highlight_git.bb
+++ b/meta-oe/recipes-support/source-highlight/source-highlight_git.bb
@@ -1,0 +1,36 @@
+SUMMARY = "Syntax highlight utility"
+DESCRIPTION = "Source-highlight converts source code to formatted text with syntax highlighting."
+HOMEPAGE = "https://www.gnu.org/software/src-highlite/"
+LICENSE = "GPL-3.0-only"
+SECTION = "libs"
+LIC_FILES_CHKSUM = "file://COPYING;;md5=ff95bfe019feaf92f524b73dd79e76eb"
+
+SRCREV = "894cacd0799ca60afa359a63782729dec76cbb79"
+PV = "3.1.9+git"
+SRC_URI = "git://git.savannah.gnu.org/git/src-highlite.git;protocol=https;branch=master"
+S = "${WORKDIR}/git"
+
+inherit autotools pkgconfig
+
+DEPENDS:append = " bison-native boost"
+
+DEPENDS:append:class-target = " ${PN}-native"
+
+EXTRA_OECONF="--with-boost-regex=boost_regex"
+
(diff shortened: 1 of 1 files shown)


Ecordonnier:
I initially send the patch to oe-core, but the response I got indicates meta-oe would be a better fit: https://lists.openembedded.org/g/openembedded-core/topic/112197845#msg214718

kraj:
applied with 285142657b

meta-oe/recipes-support/source-highlight/source-highlight_git.bb
```diff
@@ -0,0 +1,36 @@
...
+SRCREV = "894cacd0799ca60afa359a63782729dec76cbb79"
+PV = "3.1.9+git"
+SRC_URI = "git://git.savannah.gnu.org/git/src-highlite.git;protocol=https;branch=master"
+S = "${WORKDIR}/git"
+
+inherit autotools pkgconfig
+
+DEPENDS:append = " bison-native boost"
+
+DEPENDS:append:class-target = " ${PN}-native"
```
shr-project, line 17: This doesn't work well with multilib, causes:

ERROR: Nothing PROVIDES 'lib32-source-highlight-native' (but virtual:multilib:lib32:meta-oe/meta-oe/recipes-support/source-highlight/source-highlight_git.bb DEPENDS on or otherwise requires it). Close matches:
  lib32-source-highlight
  source-highlight
  source-highlight-native

meta-oe/recipes-support/source-highlight/source-highlight_git.bb
```diff
@@ -0,0 +1,36 @@
+SUMMARY = "Syntax highlight utility"
+DESCRIPTION = "Source-highlight converts source code to formatted text with syntax highlighting."
+HOMEPAGE = "https://www.gnu.org/software/src-highlite/"
+LICENSE = "GPL-3.0-only"
+SECTION = "libs"
+LIC_FILES_CHKSUM = "file://COPYING;;md5=ff95bfe019feaf92f524b73dd79e76eb"
```
shr-project, line 6: Probably doesn't cause any issue, but the ; is doubled here.

meta-oe/recipes-support/source-highlight/source-highlight_git.bb
Ecordonnier: Thanks for your feedback. I've fixed it in https://github.com/openembedded/meta-openembedded/pull/958

PR #957, opened 2025-04-12
uutils-coreutils: fix selinux PACKAGECONFIG option

Ecordonnier:
This does not compile any more after https://git.openembedded.org/openembedded-core/commit/?id=16745b20452de60ae2474433cc1a2fb1ed9f6a64 which appends the contents of PACKAGECONFIG_CONFARGS to the cargo build command.


kraj:
applied with c5a8eddd99

PR #960, opened 2025-04-17
chrony: use inherit_defer for conditional inherit of useradd

afenkart:


kraj:
applied to master @ 63df976d8e

PR #961, opened 2025-04-30
RFC: openh264: Is LICENSE_FLAGS needed?

benjamin1313:
Hey. 

For what I understand the BSD-2-Clause is an open source license, not a commercial one.
So I was wondering why I need to add it to LICENSE_FLAGS_ACCEPTED in order to use it in a project?


kraj:
There are some specific situations where additional licensing might be needed:

- If you modify OpenH264 in ways that aren't covered by Cisco's patent grant
- If you implement your own H.264 encoder/decoder instead of using OpenH264
- If you're using H.264 in ways not covered by Cisco's patent license (certain broadcast applications)

So we want the user to be aware of this and make the choice themselves

PR #962, opened 2025-05-05
Fix uhubctl_2.6.0.bb, add missing inherit

ivangalkin:
Fix error while building of uhubctl.

```
| The correct version of pkg-config is not being used!
| Make sure the recipe inherits 'pkgconfig'.
```

uhubctl depends on pkg-config in its [Makefile](https://github.com/mvp/uhubctl/blob/fbe812bd83fcd2801131a8ea2479104e0d2b2352/Makefile#L29).


kraj:
applied to master with 581819dac2

PR #971, opened 2025-06-29
cppzmq: support native build

yishai1999:
Add support for native build of cppzmq


kraj:
applied with a8dfd10600

Issue #878, opened 2024-10-06
kernel-selftest: ptp: testptp is broken in latest kernel version


cshilwant:
The latest kernels could not build ptp correctly.
For reference, with 6.12-rc1 I observe the following,

`
| make: Entering directory '/mnt/chirag-sdk/yocto-build_am62lxx-evm/build/arago-tmp-default-glibc/work/am62lxx_evm-oe-linux/kernel-selftest/1.0/kernel-selftest-1.0/tools/testing/selftests/rtc'
inux/kernel-selftest/1.0/kernel-selftest-1.0/tools/testing/selftests/ptp/testptp
| testptp.c: In function 'main':
| testptp.c:613:31: error: 'PTP_MASK_CLEAR_ALL' undeclared (first use in this function)
|   613 |                 if (ioctl(fd, PTP_MASK_CLEAR_ALL)) {
|       |                               ^~~~~~~~~~~~~~~~~~
| testptp.c:613:31: note: each undeclared identifier is reported only once for each function it appears in
| testptp.c:615:38: error: 'PTP_MASK_EN_SINGLE' undeclared (first use in this function)
|   615 |                 } else if (ioctl(fd, PTP_MASK_EN_SINGLE, (unsigned int *)&channel)) {
|       |                                      ^~~~~~~~~~~~~~~~~~
| make: *** [../lib.mk:222: /mnt/chirag-sdk/yocto-build_am62lxx-evm/build/arago-tmp-default-glibc/work/am62lxx_evm-oe-linux/kernel-selftest/1.0/kernel-selftest-1.0/tools/testing/selftests/ptp/testptp] Error 1
`

I see both the identifiers are actually defined under https://github.com/torvalds/linux/blob/v6.12-rc1/include/uapi/linux/ptp_clock.h#L246

It appears there’s a procedural problem with the kernel builds since issues like this aren’t being detected during their standard testing. It’s likely that they aren’t consistently running the self-tests.

This change was specifically introduced in v6.7 - https://github.com/torvalds/linux/commit/c5a445b1e9347b14752b01f1a304bd7a2f260acc

So, I believe testptp is broken since then.

@kraj For time being should we consider dropping ptp from https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/recipes-kernel/kernel-selftest/kernel-selftest.bb#L53 ?

kraj:
[...]

PR #975, opened 2025-07-07
kernel-selftest: Fix PTP selftest compilation for kernel 6.7+

jacobpanov:
## Summary

This PR fixes the failure to compile the PTP selftest (`testptp.c`) with newer kernel versions (6.7 and above). The issue is caused by missing header definitions like `PTP_MASK_CLEAR_ALL` and `PTP_MASK_EN_SINGLE`, which were introduced in [kernel commit c5a445b](https://github.com/torvalds/linux/commit/c5a445b).

## Details

When building the `kernel-selftest` recipe, `testptp.c` fails with the following errors:

error: 'PTP_MASK_CLEAR_ALL' undeclared
error: 'PTP_MASK_EN_SINGLE' undeclared

These definitions exist in newer versions of `ptp_clock.h`, but the build process wasn’t using the updated kernel headers.

## Fix

To fix the build, I added the proper kernel header path to the compile flags by modifying the `do_compile()` step:

export CFLAGS="${CFLAGS} -I${STAGING_KERNEL_BUILDDIR}/usr/include"

This ensures that testptp.c can access the updated definitions and compile successfully with kernel 6.7+.

Testing
Verified that the fix allows testptp.c to compile correctly using a kernel built from v6.12-rc1.

Confirmed that the kernel-selftest recipe builds cleanly in a minimal environment.

Ensured that nothing else in the recipe broke as a result of the changes.

Fixes: #878

This was my first time contributing to meta-openembedded, and it took a lot of trial and error to simulate and validate a working fix. I'm a recent graduate looking to expand my skillset by working on different projects so I can eventually be hired! Thank you!


kraj:
Please rebase to have just commit that fixes the recipe. Other commit to add/delete .devcontainer should go.

PR #976, opened 2025-07-07
fix env in recipe libinih_58.bb

Reckoner911:


meta-oe/recipes-support/inih/libinih_58.bb
```diff
@@ -10,4 +10,6 @@ SRC_URI = "git://github.com/benhoyt/inih.git;protocol=https;branch=master"
 
 SRCREV = "5cc5e2c24642513aaa5b19126aad42d0e4e0923e"
 
+S = "${WORKDIR}/git"
```
kraj, line 13: do we still need this change after S moved into UNPACKDIR ?

OldManYellsAtCloud:
Based on the error message you shared you are using Scarthgap, which is not compatible with the master branch of meta-oe anymore. Try the scarthgap branch, which is still maintained with fixes.

PR #982, opened 2025-07-22
php: fix removal of --with-zlib-dir configure option

gizahNL:
Build with zip added to packageconfig fails on walnascar, culprit is the upstream removal of the --with-zlib-dir configure option.


kraj:
applied to master with 90fa225b86fa3df35b3273a5522b72e95ae258d7

PR #983, opened 2025-08-07
perfetto: upgrade 47.0 -> 51.2

Ecordonnier:
Release Notes: https://github.com/google/perfetto/releases/tag/v51.2


kraj (changes requested):
I am seeing
```
stdio: ERROR: libperfetto-51.2-r0 do_populate_lic: QA Issue: libperfetto: The LIC_FILES_CHKSUM does not match for file://LICENSE;md5=d2572d98547d43906b53615f856a8c2d
stdio: ERROR: libperfetto-51.2-r0 do_populate_lic: Fatal QA errors were found, failing task.
```

Ecordonnier:
Sorry about that, I had tested perfetto on target, but had forgotten that perfetto doesn't depend on libperfetto which needs to be tested independently.

Ecordonnier:
I would like to send a follow-up PR adding support for "traced_perf" which is required to support "[callstack_sampling](https://perfetto.dev/docs/getting-started/cpu-profiling)". Basically you get all function-names shown in the perfetto traces with it, like a normal profiler. I have debugged this today, and the issue is that traced_perf can be compiled only with clang (see https://github.com/google/perfetto/blob/092d0ceace6fa516fac1bd4e715c226eaaebe26e/gn/perfetto.gni#L178 ). One simple way to fix this would be to have `TOOLCHAIN = "clang"` in perfetto.bb.
Is that OK in meta-openembedded, now that clang support is available directly in poky?

kraj:
> I would like to send a follow-up PR adding support for "traced_perf" which is required to support "[callstack_sampling](https://perfetto.dev/docs/getting-started/cpu-profiling)". Basically you get all function-names shown in the perfetto traces with it, like a normal profiler. I have debugged this today, and the issue is that traced_perf can be compiled only with clang (see https://github.com/google/perfetto/blob/092d0ceace6fa516fac1bd4e715c226eaaebe26e/gn/perfetto.gni#L178 ). One simple way to fix this would be to have `TOOLCHAIN = "clang"` in perfetto.bb. Is that OK in meta-openembedded, now that clang support is available directly in poky?

yes its perfectly fine.

kraj:
applied with 1b2ebce563

PR #984, opened 2025-08-09
perfetto: fix error in qemux86-64

Ecordonnier:
This fixes this error:

```
root@qemux86-64:~# tracebox --txt -c /tmp/config.pbtxt -o /tmp/perfetto-trace.pb
This executable requires a x86_64 cpu that supports SSE4.2, BMI2 and AVX2.
Rebuild with enable_perfetto_x64_cpu_opt=false.
```


kraj:
Does this work fine if you use revert this commit in poky

https://git.yoctoproject.org/poky/commit/?id=3d67b6c60fa968cd36b51c63e2a7f11bb87dd315

PR #985, opened 2025-08-09
perfetto: enabled traced_perf by compiling with clang

Ecordonnier:
traced_perf is required for using e.g. callstack_sampling, see https://perfetto.dev/docs/getting-started/cpu-profiling

Tested on qemu:

Starting perfetto using a configuration which uses `callstack_sampling` starts traced_perf as a side-effect:

```
root@qemux86-64:~#  tracebox --txt -c /tmp/config.pbtxt -o /tmp/perfetto-trace.pb &
root@qemux86-64:~# ps -eo pid,cmd | grep tracebox
  529 tracebox --txt -c /tmp/config.pbtxt -o /tmp/perfetto-trace.pb
  530 /usr/bin/tracebox traced
  533 /usr/bin/tracebox traced_probes --reset-ftrace
  536 /usr/bin/tracebox traced_perf
```


kraj:
applied with ab5cd440b3

PR #991, opened 2025-08-26
kernel-selftest: handle missing -64.h headers

nylon7:
Some toolchains ship only bits/*.h without the -64.h suffix, causing the recipe to fail. Add a fallback to use *.h if *-64.h is not found, and warn if neither exists.


kraj:
applied with 9472f4a728 to master

PR #996, opened 2025-09-09
uutils-coreutils: fix file not being packaged

Ecordonnier:


kraj:
applied with 98cdacb8ba

PR #999, opened 2025-09-11
tinyalsa: Add meson build option for plugin support

qti-sbojja:
Add patch to enable/disable plugin support for Meson build system.

Upstream-Status: Fromlist  https://github.com/tinyalsa/tinyalsa/pull/269


kraj:
applied with 4e0a5e434c

PR #1000, opened 2025-09-12
uutils-coreutils systemd support

Ecordonnier:
Commit 1:

    uutils-coreutils: upgrade to 0.2.2
    
    See https://github.com/uutils/coreutils/releases/tag/0.2.2

Commit 2:

    uutils-coreutils: add feat_systemd_logind
    
    It allows commands such as "who" to work even when systemd is compiled without "utmp" support.

### DIFF
--- a/meta-oe/recipes-core/uutils-coreutils/uutils-coreutils_0.2.2.bb
+++ b/meta-oe/recipes-core/uutils-coreutils/uutils-coreutils_0.2.2.bb
@@ -8,11 +8,11 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=e74349878141b240070458d414ab3b64"
 inherit cargo cargo-update-recipe-crates
 
 SRC_URI = " \
-    git://github.com/uutils/coreutils.git;protocol=https;branch=main \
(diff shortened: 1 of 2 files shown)


meta-oe/recipes-core/uutils-coreutils/uutils-coreutils_0.2.2.bb
```diff
@@ -7,17 +7,21 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=e74349878141b240070458d414ab3b64"
 
 inherit cargo cargo-update-recipe-crates
 
-SRC_URI = "git://github.com/uutils/coreutils.git;protocol=https;branch=main"
+SRC_URI = " \
+    git://github.com/uutils/coreutils.git;protocol=https;branch=main \
```
OldManYellsAtCloud, line 11: Since you are touching it, could you please also add a `;tag=${PV}` to the end of the git uri?

Ecordonnier: done

PR #1012, opened 2025-11-04
[meta-oe] Add gpp: a general-purpose preprocessor

b4u-mw:
This adds "gpp", a general purpose preprocessor. It's very similar to cpp's preprocessor but nor bound to the language. E.g. it allows to preprocess yaml files (with # as comment which breaks cpp preprocessor). That's very useful to process config files in recipes.


kraj:
commit message should be improved 

```
gpp: Add recipe

gpp is a general-purpose preprocessor
```

Then it should have `Signed-off-by`

kraj:
applied to master with abbbaf2ef8

PR #1016, opened 2025-11-29
Update SRC_URI to use HTTPS protocol

yzxiu:
Update SRC_URI to use the HTTPS protocol to make it easier to use http_proxy for proxying


OldManYellsAtCloud:
Isn't this a generic problem for you, with all recipes? Wouldn't a PREMIRROR, something like
`PREMIRRORS += "git://.*/.* git://HOST/PATH;protocol=https"` in `local.conf` work in a more generic way?

If not, then there are some issues with this submission, it doesn't meet the [contribution guidelines](https://docs.yoctoproject.org/dev/contributor-guide/submit-changes.html#implement-and-commit-changes). Please include the recipename in the subject, add some description in the commit why this change is necessary, and don't forget to sign it off at the end.

PR #1020, opened 2026-01-15
open62541: update to v1.4

JohannesKauffmann:
Use the latest commit from the 1.4 branch; the last 1.4 release was 3 months ago so it contains important fixes.

- The contents of /usr/share/ are slightly different, so change the path
slightly.
- The new patch fixes the .pc file generation (it also ensures that
there are no references to absolute paths in the .pc file which would
need to be removed again).
- PubSub information model is now enabled by default, add a new option
to disable it (disabling only pubsub isn't enough).

Signed-off-by: Johannes Kauffmann <johanneskauffmann@hotmail.com>


kraj:
merged to master with 3e3f0eac0f

PR #1021, opened 2026-01-21
tinycompress: install compress_ops.h as public header

mpratyus:
Add a patch to install compress_ops.h as a public header for tinycompress, and update the recipe to apply it.

This is required for users that need access to compress_ops from external components.

Upstream-Status: submitted


kraj:
installed with def8784b1d4c9ae6275de0b49f4154cc5756095d

PR #1026, opened 2026-02-09
rocksdb: support pkg-config based zstd detection

flowergom:
In Yocto builds, zstd is built via Makefile and does not install CMake package configuration files. 
As a result, Findzstd.cmake fails to detect ZSTD_INCLUDE_DIRS.
Add pkg-config based detection as a fallback to properly locate zstd headers and libraries.

fix error:
|   Could NOT find zstd (missing: ZSTD_INCLUDE_DIRS)
| Call Stack (most recent call first):

Upstream-Status: Submitted [https://github.com/facebook/rocksdb/pull/14308]


kraj:
applied with bf898320ec to master

PR #1037, opened 2026-04-01
networkmanager: enable modemmanager support for 'phone' devices

asusmith-qti:
Enable ModemManager in NetworkManager when the 'phone' MACHINE_FEATURE is enabled.


meta-networking/recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -0,0 +1 @@
+PACKAGECONFIG:append:qcom = " ${@bb.utils.contains('MACHINE_FEATURES', 'phone', ' modemmanager ', '', d)}"
```
anujm1: `:qcom` override is not needed here. bbappend is not needed for this layer. Please send patch to mailing list.

PR #1044, opened 2026-04-07
Bslawomir330 patch 1

bslawomir330:
# FRR frr-reload.py path update to /usr/libexec/frr/

## Summary
This pull request updates the FRR (FRRouting) recipe to use the correct path for `frr-reload.py` in `/usr/libexec/frr/` instead of `/usr/lib/frr/`.

## Changes
- Added new patch file: `0002-tools-update-frr-reload.py-path-to-usr-libexec-frr.patch`
  - Updates `tools/frr-reload` script to look for `frr-reload.py` in `/usr/libexec/frr/`
  - Ensures proper execution path for the FRR reload utility
- Updated `frr_9.1.3.bb` recipe to include the new patch in the build process

## Details
- **Recipe**: `meta-networking/recipes-protocols/frr/frr_9.1.3.bb`
- **FRR Version**: 9.1.3
- **Branch**: stable/9.1

## Testing
Please verify that:
- The FRR package builds successfully with this patch
- The `frr-reload.py` script is correctly located and executable in `/usr/libexec/frr/`
- Existing functionality is not affected

### DIFF
--- a/meta-networking/recipes-protocols/frr/frr/0002-tools-update-frr-reload.py-path-to-usr-libexec-frr.patch
+++ b/meta-networking/recipes-protocols/frr/frr/0002-tools-update-frr-reload.py-path-to-usr-libexec-frr.patch
@@ -1,0 +1,25 @@
+From: Sławomir Błaszczyk <bslawomir330@gmail.com>
+Date: Tue, 07 Apr 2026 13:54:00 +0200
+Subject: [PATCH] tools: update frr-reload.py path to /usr/libexec/frr/
+
+Upstream-Status: Submitted [https://github.com/openembedded/meta-openembedded/pull/1044]
+---
+ tools/frr-reload | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/tools/frr-reload b/tools/frr-reload
+index 75b31d0622..eb784da3d6 100755
+--- a/tools/frr-reload
++++ b/tools/frr-reload
+@@ -1,7 +1,7 @@
+ #!/bin/sh
+
+-if test -e /usr/lib/frr/frr-reload.py; then
+-   exec /usr/lib/frr/frr-reload.py --reload /etc/frr/frr.conf
++if test -e /usr/libexec/frr/frr-reload.py; then
++   exec /usr/frrexec/frr-reload.py --reload /etc/frr/frr.conf
+ fi
+ >&2 echo "Please install frr-pythontools package. Required for reload"
+ exit 1
+--
+2.41.0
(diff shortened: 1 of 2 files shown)


OldManYellsAtCloud:
@anujm1 ping

anujm1:
> I do not understand the claim regarding a fatal QA error.
> 
> The location of executable files in /usr/libexec/frr is appropriate; after all, it concerns scripts, not libraries.
> 
> This is not a bug in the frr-reload.sh script, 

It seems like a bug in that script to me. We pass `sbindir` to a different value and the script isn't respecting it and hardcoding a presumed value.

> but rather a misalignment with the Autotools configuration process. I believe it is a waste of time to address this at the source of the problem. Patching is intended, among other things, to correct such issues.

It's not a waste of time because such patches then need to be maintained for the entire duration of the release and are best fixed upstream when possible.

meta-networking/recipes-protocols/frr/frr/0002-tools-update-frr-reload.py-path-to-usr-libexec-frr.patch
```diff
@@ -0,0 +1,25 @@
...
+index 75b31d0622..eb784da3d6 100755
+--- a/tools/frr-reload
++++ b/tools/frr-reload
+@@ -1,7 +1,7 @@
+ #!/bin/sh
+
+-if test -e /usr/lib/frr/frr-reload.py; then
+-   exec /usr/lib/frr/frr-reload.py --reload /etc/frr/frr.conf
++if test -e /usr/libexec/frr/frr-reload.py; then
++   exec /usr/frrexec/frr-reload.py --reload /etc/frr/frr.conf
```
anujm1, line 20: It doesn't look like this has been tested.

meta-networking/recipes-protocols/frr/frr/0002-tools-update-frr-reload.py-path-to-usr-libexec-frr.patch
```diff
@@ -0,0 +1,25 @@
+From: Sławomir Błaszczyk <bslawomir330@gmail.com>
+Date: Tue, 07 Apr 2026 13:54:00 +0200
+Subject: [PATCH] tools: update frr-reload.py path to /usr/libexec/frr/
+
+Upstream-Status: Submitted [https://github.com/openembedded/meta-openembedded/pull/1044]
```
anujm1, line 5: Submitted is for patches submitted upstream. This is a valid issue and the status should be Pending. At the very least, it should be reported upstream. This would also need your Signed-off-by.

anujm1:
Kindly also note that the preferred way to submit patches is through mailing list so others can also chime in.

PR #1045, opened 2026-04-08
Revert "Selinux policy for adb service"

GargiQcom:
As per our discussion with refpolicy maintainer: https://github.com/SELinuxProject/refpolicy/issues/1100, initrc_t is also unconfined and there is no requirement from moving to unconfined_t.
Apart from that, on adding rules to adb shell we are seeing mount failure due to role transition, allowing which causes security concern.
Hence, we need to revert changes.


GargiQcom:
Hi @kraj 
A gentle reminder to review.

kraj:
> Hi @kraj A gentle reminder to review.

its staged in master-next

kraj:
merged with dd96b8a81e on master.

PR #1048, opened 2026-04-14
CoreMark: Add CoreMark to benchmark CPU performance

Ashwin-Prabhakar:
Add support for CoreMark benchmark in recipes-benchmark. This package can be used to benchmark CPU performance in SOC's Application processor core.


meta-oe/recipes-benchmark/coremark/coremark_git.bb
```diff
@@ -0,0 +1,23 @@
...
+do_compile() {
+    oe_runmake compile
+}
+
+do_install() {
+    install -d ${D}${bindir}
+    install -m 0755 ${S}/coremark.exe ${D}${bindir}/coremark
+}
+
+INSANE_SKIP:${PN} += "ldflags"
```
kraj: why is this needed? Can you try adding

`TARGET_CC_ARCH += "${LDFLAGS}"` and see if that helps?

OldManYellsAtCloud: Instead of skipping this QA check, you could try adding `LFLAGS_END="${LDFLAGS}"` or `XCFLAGS="${LDFLAGS}"` at the end of `oe_runmake` command - I think one of these should do the trick

meta-oe/recipes-benchmark/coremark/coremark_git.bb
```diff
@@ -0,0 +1,23 @@
+SUMMARY = "EEMBC CoreMark CPU benchmark"
+DESCRIPTION = "CoreMark is a CPU benchmark program"
+HOMEPAGE = "https://www.eembc.org/coremark/"
+LICENSE = "Apache-2.0"
```
OldManYellsAtCloud: Is it up to the user to choose Apache-2.0 to cover the whole codebase? The license seems to have some custom additions on top of Apache, though I didn't look very deep to see the actual impact of it, if it can be ignored. If users can choose Apache only, and the custom part can be ignored, maybe it would be useful to mention it in the commit message. (Otherwise it might be necessary to import the custom license text in the layer... which is simple, but still an extra step)

OldManYellsAtCloud (changes requested):

meta-oe/recipes-benchmark/coremark/coremark_git.bb
OldManYellsAtCloud: Ahhh, now I see that I was too slow with this comment.

Ashwin-Prabhakar: @kraj Thanks and this did the trick

Ashwin-Prabhakar: @OldManYellsAtCloud, Thanks for the inputs. Integrating @kraj suggestion fixed the QA failures.

meta-oe/recipes-benchmark/coremark/coremark_git.bb
Ashwin-Prabhakar: @OldManYellsAtCloud  from what I understood the software can be used as is under apache2.0 or can be used under a closed license by enabling the commercial flag. But I am not sure myself and I was interested to use this under the Apache-2.0 license.

kraj: I think to be more accurate it could be 

`LICENSE = "Apache-2.0 & LicenseRef-EEMBC-AUA"`

Where AUA is the EEMBC Acceptable Use Agreement and be aware of the AUA restrictions.

Ashwin-Prabhakar: Thanks @kraj , I have update the License as per advise. Do you also want me to create a custom License in the /licenses directory and copy the license terms there ?

kraj:
applied with 6b29351885 to master

PR #1051, opened 2026-04-29
cloudflared: add initial recipe for 2026.3.0 version

embetrix:
Add initial recipe for Cloudflare Tunnel client (cloudflared). The upstream source vendors all Go dependencies so no go-mods.inc is needed.
Includes systemd service with token-based authentication via /etc/default/cloudflared.


embetrix:
sent to mailing list so close it here.

PR #1053, opened 2026-05-27
libuio: fix FILE descriptor leak

Qliangw:
The function uio_line_from_file() fails to close the FILE pointer when fgets() returns NULL, causing a file descriptor leak.

This can be triggered when reading from /sys files that return empty content, leading to resource exhaustion over time.

Fix this by using goto-based error handling to ensure fclose() is called on all exit paths.

### DIFF
--- a/meta-oe/recipes-extended/libuio/libuio/fix-fclose-leak-in-uio_line_from_file.patch
+++ b/meta-oe/recipes-extended/libuio/libuio/fix-fclose-leak-in-uio_line_from_file.patch
@@ -1,0 +1,31 @@
+From: Jason Smith <qili00001@gmail.com>
+Subject: [PATCH] Fix FILE descriptor leak in uio_line_from_file()
+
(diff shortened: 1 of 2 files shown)


kraj (changes requested):
Please submit this patches upstream to libuio for review and comment and add Upstream-Status: tag to the patch header.

Secondly, apply the patch directly in bb file no need to do it via a bbappend.

Qliangw:
> Please submit this patches upstream to libuio for review and comment and add Upstream-Status: tag to the patch header.
> 
> Secondly, apply the patch directly in bb file no need to do it via a bbappend.


Done.

1. The patch has been submitted upstream to [libuio Merge Request #2](https://sourceforge.net/p/libuio/code/merge-requests/2/) for review and comments
2. The `Upstream-Status` tag has been added to the patch header
3. The patch is now applied directly in the `.bb` file, and the `.bbappend` has been removed

kraj:
applied with cd75edf25d to `master`

PR #1054, opened 2026-06-02
osbench: Add osbench recipe for OS benchmarking tools

Ashwin-Prabhakar:
Add a new recipe for osbench, a collection of micro-benchmarks to measure operating system primitives such as process creation, thread creation, file operations, and memory allocation.


kraj:
applied with 49dac62f88 to master

PR #1055, opened 2026-06-02
ramspeed: Add recipe for cache and memory benchmarking tool

Ashwin-Prabhakar:
Add a new recipe for ramspeed, a lightweight benchmark utility that measures cache and memory bandwidth using a set of synthetic tests.

### DIFF
--- a/meta-oe/licenses/Alasir
+++ b/meta-oe/licenses/Alasir
@@ -1,0 +1,47 @@
+                              The Alasir Licence
+
+
+    This is a free software. It's provided as-is and carries absolutely no
+warranty or responsibility by the author and the contributors, neither in
+general nor in particular. No matter if this software is able or unable to
+cause any damage to your or third party's computer hardware, software, or any
+other asset available, neither the author nor a separate contributor may be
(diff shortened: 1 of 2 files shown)


meta-oe/licenses/Alasir
```diff
@@ -0,0 +1,47 @@
...
+copyright itself indicates the rights of ownership over the code contributed.
+Back and forth, the author is defined as the one who holds the oldest
+copyright over the software. Furthermore, the software is defined as either
+source or binary computer code, which is organised in the form of a single
+computer file usually.
+
+    The software (the whole or a part of it) is prohibited from being sold or
+leased in any form or manner with the only possible exceptions:
+
+a) money may be charged for a physical medium used to transfer the software;
```
priv-kweihmann, line 26: The license seems to prohibit commercial use without an additional license agreement.
So either LICENSE_FLAGS_ACCEPTED should be used to prevent users to accidentally running into legal issues (or any other way) or the recipe shouldn't be accepted into meta-oe

kraj:
applied with fd914c8d8f

PR #1056, opened 2026-06-09
crash 8.0.6 -> 9.0.2 + crash-memory-driver

RobertBerger:
Updated crash recipe from 8.0.6 to 9.0.2

- Slightly different approach, which seems to be a bit less hacky
- Some patches were removed since they seem to be obsolete now
- `crash-native` builds and runs
- `crash-cross-canadian-aarch64` builds—I did not build an SDK to see if it runs.
- `crash` for aarch64 target builds and runs

```
# target binary:
crash --buildinfo
build_command: crash
build_data: reproducible build
build_target: ARM64
build_version: 9.0.2
compiler version: aarch64-resy-linux-gcc (GCC) 15.2.0
```

```
# crash-native:
crash-native/9.0.2/recipe-sysroot-native/usr/bin/crash --buildinfo
build_command: crash
build_data: reproducible build
build_target: X86_64
build_version: 9.0.2
compiler version: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
```

Added recipe for crash memory driver
- [memory driver readme](https://github.com/crash-utility/crash/tree/master/memory_driver)


kraj:
I think its using `0001-hacked-Makefilefor-module.bbclass.patch` but this patch is missing. Its failing as a result see

https://autobuilder.yoctoproject.org/valkyrie/api/v2/logs/7106116/raw_inline

meta-oe/recipes-kernel/crash/crash-memory-driver/0001-hacked-Makefile-for-module.bbclass.patch
```diff
@@ -0,0 +1,57 @@
...
+@@ -1,21 +1,19 @@
+-# This program is free software; you can redistribute it and/or modify
+-# it under the terms of the GNU General Public License as published by
+-# the Free Software Foundation; either version 2 of the License, or
+-# (at your option) any later version.
+-#
+-# This program is distributed in the hope that it will be useful,
+-# but WITHOUT ANY WARRANTY; without even the implied warranty of
+-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-# GNU General Public License for more details.
```
OldManYellsAtCloud: Can we just leave the license part as it is?

meta-oe/recipes-kernel/crash/crash/0003-Fix-build-failure-in-readline-lib.patch
OldManYellsAtCloud, line 1: Looking at the referenced commit (https://github.com/crash-utility/crash/commit/772fbb1), it seems to be included starting version 9.0.0 - though it wasn't included in the new patch for gdb 16.2.
Is this patch still needed?
I got the same question for `0005-Fix-build-failure-on-32bit-machine-i686.patch` file.

meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb
```diff
@@ -0,0 +1,30 @@
+DESCRIPTION = "Once installed, the /dev/crash driver will be used by default for live system crash sessions."
+SECTION = "devel"
+SUMMARY = "/dev/crash driver"
+HOMEPAGE = "https://github.com/crash-utility/crash/tree/master/memory_driver"
+RECIPE_MAINTAINER = "Robert Berger <Robert.Berger@ReliableEmbeddedSystems.com>"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0-only;md5=801f80980d171dd6425610833a22dbe6"
```
OldManYellsAtCloud: Not a very strong opinion, but I think referencing the source file's license declaration would be better: https://github.com/crash-utility/crash/blob/master/memory_driver/crash.c#L9
Also, it seems to be gpl-2-or-later, not `only`

meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb
```diff
@@ -0,0 +1,30 @@
...
+
+inherit module
+
+SRCREV = "61fe107ff96a22e7df0029877529b7ce6da36850"
+SRC_URI = "git://github.com/crash-utility/crash;protocol=https;branch=master;subpath=memory_driver;destsuffix=${BP} \
+           file://0001-hacked-Makefilefor-module.bbclass.patch;striplevel=2 \
+           file://0002-add-sparse-support.patch;striplevel=2 \
+"
+PV = "9.0.2+git${SRCPV}"
```
OldManYellsAtCloud: SRCPV is not needed anymore, it can be dropped

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
-DESCRIPTION = "The core analysis suite is a self-contained tool that can be used to\
-investigate either live systems, kernel core dumps created from the\
-netdump, diskdump and kdump packages from Red Hat Linux, the mcore kernel patch\
-offered by Mission Critical Linux, or the LKCD kernel patch."
-
-HOMEPAGE = "http://people.redhat.com/anderson"
+SUMMARY = "Kernel analysis utility for live systems and core dumpfiles"
+DESCRIPTION = "The core analysis suite is a self-contained tool that can be used to \
+investigate either live systems, kernel core dumps created from kdump, or mcore dumpfiles."
+HOMEPAGE = "https://github.com"
```
OldManYellsAtCloud: Probably https://github.com/crash-utility/crash/ or https://crash-utility.github.io/ would be better for homepage

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
 SECTION = "devel"
 
-LICENSE = "GPL-3.0-only"
-LIC_FILES_CHKSUM = "file://COPYING3;md5=d32239bcb673463ab874e80d47fae504"
-
-DEPENDS = "zlib readline coreutils-native ncurses-native"
+LICENSE = "GPL-3.0-or-later"
```
OldManYellsAtCloud: Is this really `or-later`? To me it looks `-only`, though I might be missing something.

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
+# to get the right compiler info into ./crash --buildinfo
+
+SRCREV = "61fe107ff96a22e7df0029877529b7ce6da36850"
+#PV = "9.0.2"
+SRC_URI[gdb.sha256sum] = "bdc1da4a033280ac752e7d34b0418efaa45bed093235cb88e62ea961752a37f8"
 
-SRC_URI[gdb.sha256sum] = "b33ad58d687487a821ec8d878daab0f716be60d0936f2e3ac5cf08419ce70350"
 
-UPSTREAM_CHECK_URI = "https://github.com/crash-utility/crash/releases"
+UPSTREAM_CHECK_URI = "https://github.com/releases"
```
OldManYellsAtCloud: This seems to be a typo or copy-paste error?

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
+TARGET_CFLAGS:append = " -fcommon -fpermissive"
+TARGET_CXXFLAGS:append = " -fcommon -fpermissive"
+
+export CC
+export CXX
+export CFLAGS
+export LDFLAGS
+export AR
+export LD
+export RANLIB
```
OldManYellsAtCloud: Are these exports needed explicitly? Most (all?) of these are exported by default. Or am I missing some context?

meta-oe/recipes-kernel/crash/crash_9.0.2.bb
```diff
@@ -28,3 +27,23 @@ do_install:class-cross () {
...
+
+# -->
+# crash-native:
+# crash-native/9.0.2/recipe-sysroot-native/usr/bin/crash --buildinfo
+#  build_command: crash
+#  build_data: reproducible build
+#  build_target: X86_64
+#  build_version: 9.0.2
+#  compiler version: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
+# <--
```
OldManYellsAtCloud: Was this comment left here accidentally? Or is it useful for something?

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -88,9 +141,7 @@ do_install:prepend () {
...
-# jenkins  15161  100 12.5 10389596 10321284 ?   R    11:40  28:17 /home/jenkins/oe/world/shr-core/tmp-glibc/sysroots/x86_64-linux/usr/libexec/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/4.9.2/cc1 -quiet -I . -I . -I ./common -I ./config -I ./../include/opcode -I ./../opcodes/.. -I ./../readline/.. -I ../bfd -I ./../bfd -I ./../include -I ../libdecnumber -I ./../libdecnumber -I ./gnulib/import -I build-gnulib/import -isysroot /home/jenkins/oe/world/shr-core/tmp-glibc/sysroots/qemuarm -MMD eval.d -MF .deps/eval.Tpo -MP -MT eval.o -D LOCALEDIR="/usr/local/share/locale" -D CRASH_MERGE -D HAVE_CONFIG_H -D TUI=1 eval.c -quiet -dumpbase eval.c -march=armv5te -mthumb -mthumb-interwork -mtls-dialect=gnu -auxbase-strip eval.o -g -O2 -Wall -Wpointer-arith -Wformat-nonliteral -Wno-pointer-sign -Wno-unused -Wunused-value -Wunused-function -Wno-switch -Wno-char-subscripts -Wmissing-prototypes -Wdeclaration-after-statement -Wempty-body -feliminate-unused-debug-types -o -
 ARM_INSTRUCTION_SET = "arm"
 
-# http://errors.yoctoproject.org/Errors/Details/186964/
-COMPATIBLE_HOST:libc-musl = 'null'
+ERROR_QA:remove = "buildpaths"
+WARN_QA:append = " buildpaths"
```
OldManYellsAtCloud, line 141: If this is really unavoidable, could you please add some notes about it in the commit message (or even as a comment in the recipe), why this is needed?

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
+DEPENDS:append = " coreutils-native bison-native flex-native"
 
-SRC_URI = "git://github.com/crash-utility/${BPN}.git;branch=master;protocol=https \
-           ${GNU_MIRROR}/gdb/gdb-10.2.tar.gz;name=gdb;subdir=${BB_GIT_DEFAULT_DESTSUFFIX} \
-           file://7001force_define_architecture.patch \
+SRC_URI = "git://github.com/crash-utility/${BPN}.git;branch=master;protocol=https;destsuffix=${BP} \
+           ${GNU_MIRROR}/gdb/gdb-16.2.tar.gz;name=gdb;subdir=gdb \
            file://7003cross_ranlib.patch \
-           file://0001-cross_add_configure_option.patch \
-           file://donnot-extract-gdb-during-do-compile.patch \
```
OldManYellsAtCloud, line 22: Not all of these patch files were removed from the tree (e.g. this `donnot...patch`) - could you please also drop all the files that are not needed anymore?

meta-oe/recipes-kernel/crash/crash.inc
```diff
@@ -1,80 +1,133 @@
...
+
+    # 4. Apply your original configuration text manipulations natively
+    sed -i s/FORCE_DEFINE_ARCH/"${ARCH_UPPER}"/g ${S}/configure.c
     sed -i -e 's/#define TARGET_CFLAGS_ARM_ON_X86_64.*/#define TARGET_CFLAGS_ARM_ON_X86_64\t\"TARGET_CFLAGS=-D_FILE_OFFSET_BITS=64\"/g' ${S}/configure.c
     sed -i -e 's/#define TARGET_CFLAGS_MIPS_ON_X86_64.*/#define TARGET_CFLAGS_MIPS_ON_X86_64\t\"TARGET_CFLAGS=-D_FILE_OFFSET_BITS=64\"/g' ${S}/configure.c
     sed -i 's/&gt;/>/g' ${S}/Makefile
+
+    # 5. Return to the active workspace build root to merge downstream additions
+    cd ${B}
+    if [ -f "gdb-16.2.patch" ]; then
```
OldManYellsAtCloud: I think this is overly defensive, and this `if` could be dropped: If upstream renames this patch after an update, it is easy to miss, because the lack of patch is swallowed silently. I think we should complain loud if this file doesn't exist, so the recipe can be updated also.

OldManYellsAtCloud (changes requested):
I have added a couple notes and questions, mostly small nitpicks.

meta-oe/recipes-kernel/crash/crash-memory-driver/0001-hacked-Makefile-for-module.bbclass.patch
RobertBerger: Yes, we can keep the license of the Makefile: https://github.com/openembedded/meta-openembedded/commit/7998a781f9846002975c22a2119f1bb30348a814

meta-oe/recipes-kernel/crash/crash/0003-Fix-build-failure-in-readline-lib.patch
RobertBerger: To be honest, I have no idea. I kept the patches, which were cleanly applied.

meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb
RobertBerger: I can reference the source file's license declaration. It looks like `gpl-2-or-later`, but the last line says `MODULE_LICENSE("GPL");` which means it's `GPL-2.0-only`. This is needed if a kernel module needs to have access to all kernel symbols.

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: https://github.com/crash-utility/crash/blob/master/COPYING3#L637 
```
This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
```
3 or later?

OldManYellsAtCloud: Careful there: this snippet is after the `END OF TERMS AND CONDITIONS` indicator, and not part of the actual license. It's part of the "how to use this license tutorial".

meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb
OldManYellsAtCloud: Haha, yeah, ambiguous license is always great.                 
Personally I consider the license's text itself to be stronger than the macro (I believe the kernel also only cares about being compatible with gpl2, though I might have dreamed that up).
I won't die on this hill though.

RobertBerger: note to myself: done locally

meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb
RobertBerger: note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: I leave it at or-later, but if someone wants it, I can change it to -only.

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: note to myself: done locally (removed all the exports not just line 73 to 76. My test builds went through without the exports.)

meta-oe/recipes-kernel/crash/crash_9.0.2.bb
RobertBerger: It proves that 0001-Use-CC-env-var-to-get-compiler-version.patch is applied and so the right compiler is being picked up. But I can remove the comments.
note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: I added the following:
```
# WARNING: crash-9.0.2-r0 do_package_qa: QA Issue: File /usr/bin/crash in package crash contains reference to TMPDIR [buildpaths]
#
# WARNING: crash-cross-canadian-aarch64-9.0.2-r0 do_package_qa: QA Issue: File /opt/phytec-ampliphy-resy-systemd/6.0.98-devel/sysroots/x86_64-resysdk-linux/usr/bin/aarch64-resy-linux/crash in package crash-cross-canadian-aarch64 contains reference to TMPDIR [buildpaths]
#
ERROR_QA:remove = "buildpaths"
WARN_QA:append = " buildpaths"
```
note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: I am not sure if they are needed or not. We will see once people start using the new crash version. That's the reason I left them there. But sure, I can remove them for now and keep only the ones that I use.
note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: note to myself: done locally

meta-oe/recipes-kernel/crash/crash.inc
RobertBerger: Hmm. It's getting even better: crash (which is statically linked against gdb) is probably gpl3-only, but gdb is gpl-3-or-later.  The combined work is GPL-3.0-only, so I'll change that.
```
crash 9.0.2
Copyright (C) 2002-2026  Red Hat, Inc.
Copyright (C) 2004, 2005, 2006, 2010  IBM Corporation
Copyright (C) 1999-2006  Hewlett-Packard Co
Copyright (C) 2005, 2006, 2011, 2012  Fujitsu Limited
Copyright (C) 2006, 2007  VA Linux Systems Japan K.K.
Copyright (C) 2005, 2011, 2020-2024  NEC Corporation
Copyright (C) 1999, 2002, 2007  Silicon Graphics, Inc.
Copyright (C) 1999, 2000, 2001, 2002  Mission Critical Linux, Inc.
Copyright (C) 2015, 2021  VMware, Inc.
This program is free software, covered by the GNU General Public License,
and you are welcome to change it and/or distribute copies of it under
certain conditions.  Enter "help copying" to see the conditions.
This program has absolutely no warranty.  Enter "help warranty" for details.

GNU gdb (GDB) 16.2
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

meta-oe/recipes-kernel/crash/crash/0003-Fix-build-failure-in-readline-lib.patch
OldManYellsAtCloud: Looking at it closer, both of these are dangling patches, neither of them are used in the recipes. They can be dropped.

meta-oe/recipes-kernel/crash/crash.inc
OldManYellsAtCloud: If there are 2 (or more) governing licenses, then that can be just indicated as such with the `&` operator in the `LICENSE` variable, like https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/recipes-kernel/trace-cmd/trace-cmd_3.1.5.bb#L4

If the previous license has been always wrong in the recipe, that's fair, and needs to be corrected, but it needs to be supported also with some info, usually either with explaining why the license's SPDX id is incorrect in the recipe, or showing that upstream has actually changed the license.

meta-oe/recipes-kernel/crash/crash.inc
OldManYellsAtCloud: I mean more like, why does this error happen now? Can it be corrected, instead of being ignored? It happens with some recipes that there is no choice but to ignore the QA error, but in case of this recipe this is a regression, it wasn't needed so far. The real question is, what happened that made this needed?

As a sidenote, I suspect the `--with-sysroot=${STAGING_DIR_TARGET}` is causing this in EXTRA_OEMAKE. Should this rather be `--with-build-sysroot=${STAGING_DIR_TARGET}`? (This is mostly a blind guess after taking a quick glance at the config file and the error. It can be totally wrong.)

kraj:
@RobertBerger please squash all changes into one commit.

RobertBerger:
> @RobertBerger please squash all changes into one commit.

I tried to squash it all into a single commit. Please check.

ricardosalveti:
I'm trying this new revision to unblock build with latest oe-core, but I'm getting the following:

`/build/tmp/work/armv8-2a-qcom-linux/crash/9.0.2/temp/run.do_compile.1355814: 151: pushd: not found`

Which happens with standard ubuntu, as it uses dash by default.

ricardosalveti:
Also getting:

`WARNING: crash-9.0.2-r0 do_package_qa: QA Issue: File /usr/bin/crash in package crash contains reference to TMPDIR [buildpaths]`

RobertBerger:
> Also getting:
> 
> `WARNING: crash-9.0.2-r0 do_package_qa: QA Issue: File /usr/bin/crash in package crash contains reference to TMPDIR [buildpaths]`

Yes, I know about this, but I don't know how to fix it. So I made a WARNING out of the QA warning.

RobertBerger:
> I'm trying this new revision to unblock build with latest oe-core, but I'm getting the following:
> 
> `/build/tmp/work/armv8-2a-qcom-linux/crash/9.0.2/temp/run.do_compile.1355814: 151: pushd: not found`
> 
> Which happens with standard ubuntu, as it uses dash by default.

OK, I use Bash with Ubuntu, but you are right. I guess I can fix this.

kraj:
another issue I am seeing is this

```
ERROR: Nothing PROVIDES 'sparse-native' (but /srv/pokybuild/yocto-worker/meta-oe/build/meta-openembedded/meta-oe/recipes-kernel/crash/crash-memory-driver_9.0.2.bb DEPENDS on or otherwise requires it). Close matches:
  nspr-native
  parted-native
  pcre-native
ERROR: Required build target 'meta-world-pkgdata' has no buildable providers.
Missing or unbuildable dependency chain was: ['meta-world-pkgdata', 'crash-memory-driver', 'sparse-native']

Summary: There were 2 ERROR messages, returning a non-zero exit code.


INFO: test_world_inherit_class (common.CommonCheckLayer.test_world_inherit_class)
INFO:  ... FAIL
INFO: Traceback (most recent call last):

```

RobertBerger:
@kraj Yes, you are absolutely right that there is no sparse recipe available. I have it on one of my layers. So, I guess the quickest fix would be to drop sparse support for now.

kraj:
Patches are on oe-devel ml https://patchwork.yoctoproject.org/project/oe/list/?series=48120

kraj:
This is applied to master with 496ce81934

PR #1060, opened 2026-07-24
crow: add PACKAGECONFIG for Boost support

panekk:
Add boost PACKAGECONFIG option to build against Boost.Asio. By default, keep using the standalone asio library.


kraj:
Please rebase it on top of latest master and upload PR again. Some part of this has been applied.

panekk:
@kraj With the updated master, this is no longer an update of crow, but the other changes are still useful. I updated the commit message and description.

PR #1065, opened 2026-08-23
libfaketime: extend to native

krant:
It is needed for reproducible UEFI capsules in [meta-tegra](https://github.com/OE4T/meta-tegra).


kraj:
fixed with similar patch - https://git.openembedded.org/meta-openembedded/commit/?id=54d51c2c2a2dd9ac876d469af2af2782b4caa440

krant:
Thank you!

PR #1068, opened 2026-08-31
floret: add recipe for floret 0.10.5

varunsinghal29:
Add a new recipe for floret, a fastText fork with Bloom embeddings for compact, full-coverage vectors compatible with spaCy. The recipe builds floret as a shared library, static library, and CLI binary using CMake, with zlib as a dependency.


meta-oe/recipes-support/floret/floret_0.10.5.bb
```diff
@@ -0,0 +1,99 @@
...
+
+FILES:${PN}-cli       = "${bindir}/floret"
+FILES:${PN}-dev      += "${libdir}/libfloret.so ${includedir}/floret ${libdir}/pkgconfig/floret.pc"
+FILES:${PN}-staticdev = "${libdir}/libfloret.a"
+FILES:${PN}          += "${libdir}/libfloret.so.*"
+
+# Skip buildpaths/dev-elf QA on this recipe
+# libs reference TMPDIR / are non-symlink .so
+INSANE_SKIP:${PN}           += "buildpaths"
+INSANE_SKIP:${PN}-dev       += "buildpaths dev-elf"
```
kraj, line 92: dev-elf is not needed with above fix.

meta-oe/recipes-support/floret/floret_0.10.5.bb
```diff
@@ -0,0 +1,99 @@
...
+    install -d ${D}${includedir}/floret
+    install -d ${D}${bindir}
+    install -d ${D}${libdir}/pkgconfig
+
+    # Install shared library with versioned soname
+    if [ -f ${B}/libfloret.so ]; then
+        install -m 0755 ${B}/libfloret.so ${D}${libdir}/libfloret.so.${PV}
+        ln -sf libfloret.so.${PV} ${D}${libdir}/libfloret.so.1
+        ln -sf libfloret.so.1     ${D}${libdir}/libfloret.so
+    fi
```
kraj, line 45: This does not work.

 CMake builds libfloret.so.<soversion> and puts a libfloret.so symlink
next to it. Install the real library under its SONAME and recreate the
symlink: installing through the symlink instead would ship the same
library twice and leave ${libdir}/libfloret.so as an ELF file rather than
a link, which makes debian.bbclass rename ${PN}-dev to libfloret0 as
well as ${PN} - two packages with one name, which do_create_package_spdx
then dies on. you need something like below.

```bitbake
if [ -e ${B}/libfloret.so ]; then
    soname=$(${OBJDUMP} -p ${B}/libfloret.so | awk '/SONAME/ { print $2 }')
    install -m 0755 ${B}/libfloret.so ${D}${libdir}/$soname
    ln -sf $soname ${D}${libdir}/libfloret.so
fi

```

kraj (changes requested):
Please fix the error

https://errors.yoctoproject.org/Errors/Details/968023/

meta-oe/recipes-support/floret/floret_0.10.5.bb
varunsinghal29: @kraj
The issue has been resolved. Kindly review the PR and share your feedback.

varunsinghal29:
Hi @kraj 
The reported issues have been addressed. Thanks!

Could you please review this PR again?
Thanks.

meta-oe/recipes-support/floret/files/0001-PATCH-floret-Adjust-CMake-flags-for-cross-compilation.patch
```diff
@@ -0,0 +1,42 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Varun Singhal <varusing@qti.qualcomm.com>
+Date: Tue, 27 Jan 2026 21:37:55 +0530
+Subject: [PATCH] floret: Adjust CMake flags for cross-compilation
+
+- Update cmake_minimum_required from 2.8.9 to 3.5
+- Add missing #include <cstdint> to args.cc for GCC 15.x
+
+Upstream-Status: Inappropriate [cross-compilation fix for Yocto/OE]
```
anujm1, line 9: This shouldn't be Inappropriate and OE specific and should also be submitted upstream.

meta-oe/recipes-support/floret/floret_0.10.5.bb
```diff
@@ -0,0 +1,96 @@
...
+
+LICENSE          = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=4ce16d1d2b4c556959d654417007b713"
+
+DEPENDS = "zlib"
+
+SRC_URI = "git://github.com/explosion/floret.git;protocol=https;branch=main \
+           file://0001-PATCH-floret-Adjust-CMake-flags-for-cross-compilation.patch"
+
+SRCREV = "${AUTOREV}"
```
anujm1, line 21: It can't be AUTOREV

meta-oe/recipes-support/floret/floret_0.10.5.bb
```diff
@@ -0,0 +1,96 @@
...
+exec_prefix=\${prefix}
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+
+Name: floret
+Description: fastText + Bloom embeddings for compact, full-coverage vectors
+Version: ${PV}
+Libs: -L\${libdir} -lfloret
+Cflags: -I\${includedir}/floret
+EOF
```
anujm1, line 78: What is wrong with install target and the pc file installed by project?

meta-oe/recipes-support/floret/floret_0.10.5.bb
```diff
@@ -0,0 +1,96 @@
...
+
+FILES:${PN}-cli       = "${bindir}/floret"
+FILES:${PN}-dev      += "${libdir}/libfloret.so ${includedir}/floret ${libdir}/pkgconfig/floret.pc"
+FILES:${PN}-staticdev = "${libdir}/libfloret.a"
+FILES:${PN}          += "${libdir}/libfloret.so.*"
+
+INSANE_SKIP:${PN}           += "buildpaths"
+INSANE_SKIP:${PN}-dev       += "buildpaths"
+INSANE_SKIP:${PN}-staticdev += "buildpaths"
+INSANE_SKIP:${PN}-cli       += "buildpaths"
```
anujm1, line 91: Why?

anujm1 (commented):
Please also consider submitting this through mailing list instead.

meta-oe/recipes-support/floret/floret_0.10.5.bb
kraj: good catch, my builds do not catch this yet, perhaps something to improve upon.

PR #1070, opened 2026-09-15
Add util-linux-native to Breakpad bbclass

hellow554:
without this I get the following error:

```
| DEBUG: Executing shell function breakpad_package_preprocess
| Failed to write symbol file.
| /xxx/bp-dummy/1.0/temp/run.breakpad_package_preprocess.19115: 146: rev: not found
| /xxx/bp-dummy/1.0/temp/run.breakpad_package_preprocess.19115: 146: rev: not found
| WARNING: exit code 127 from a shell command.
| DEBUG: Python function do_package finished
```

not sure if this is the proper way to fix this, but it works for me™ :)

A backport would be nice as well, thanks!


kraj:
applied to master @ c3e7da7b9a

